PR #31, opened 2015-01-05
getlantern/lantern#2134 Using SIGINT to terminate flashlight server so t...

oxtoacart:
...hat CPU profile can get written to disk


myleshorton:
LGTM

PR #32, opened 2015-01-07
getlantern/lantern#2117 Added restart of flashlight when load average ex...

oxtoacart:
...ceeds 90%

Note - this is already deployed to all flashlight servers.


salt/flashlight/flashlight.conf
```diff
@@ -33,4 +33,4 @@ chdir /home/lantern
 limit nofile 1024768 1024768
 
 # Start the process, piping output prepended with a prefix to syslog
-exec /usr/bin/flashlight -addr :62443 -role server -server "{{ grains['id'] }}.{{ zone }}" -instanceid "{{ grains['id'] }}" -statsperiod 300 -cpuprofile cpu.prof 2>&1 | logger -t flashlight
\ No newline at end of file
+exec /usr/bin/flashlight -addr :62443 -role server -server "{{ grains['id'] }}.{{ zone }}" -instanceid "{{ grains['id'] }}" -statsperiod 300 -memprofile mem.prof 2>&1 | logger -t flashlight
```
oxtoacart: Since the cpu spike issue is fixed, I switched to memory profiling.  This does add some overhead, but it doesn't seem to be too bad.  I logged https://github.com/getlantern/lantern/issues/2145 to remove this when we feel we don't need it anymore.

myleshorton:
We could theoretically add in a droplet resize here as well...

oxtoacart:
> We could theoretically add in a droplet resize here as well...

Yup!

oxtoacart:
Per most recent discussion, rather than a droplet resize, we'd probably want to just destroy the droplet.

aranhoide:
LGTM, manually tested it in an instance by using boom to bring the 15m load average to >90% and running `monitor.bash` from the shell.

oxtoacart:
Thanks!

PR #33, opened 2015-01-08
getlantern/lantern#2148 Upgraded servers to flashlight 0.0.6 to include ...

oxtoacart:
...new stats for distinctClients and distinctDestHosts


aranhoide:
LGTM.  I cheated a bit on the QA front by just checking that the release exists at that URL, which is the only thing that could conceivably fail here.

oxtoacart:
@aranhoide Thanks!

PR #34, opened 2015-01-19
Added support for Peerscanner

oxtoacart:


bin/fake_controller.py
```diff
@@ -67,6 +70,8 @@ def print_usage():
             launch_fl(*sys.argv[2:])
         elif cmd == 'launch-wd':
             launch_wd(*sys.argv[2:])
+        elif cmd == 'launch-ps':
+            launch_ps(*sys.argv[2:])
```
aranhoide: Updating the usage message would be nice too.  I'll do that when merging the azure branch, since IIRC I did a bit of cleanup there.

oxtoacart: Thanks!

salt/flashlight/flashlight.conf
```diff
@@ -33,4 +33,4 @@ chdir /home/lantern
 limit nofile 1024768 1024768
 
 # Start the process, piping output prepended with a prefix to syslog
-exec /usr/bin/flashlight -addr :62443 -role server -server "{{ grains['id'] }}.{{ zone }}" -instanceid "{{ grains['id'] }}" -statsperiod 300 2>&1 | logger -t flashlight
\ No newline at end of file
+exec /usr/bin/flashlight -addr :62443 -role server -server "{{ grains['id'] }}.{{ zone }}" -instanceid "{{ grains['id'] }}" -statsperiod 300 -registerat https://peerscanner.getiantem.org 2>&1 | logger -t flashlight
```
aranhoide: As per our chat, ideally we would have a testing round robin (which perhaps means a testing peerscanner).  This can be added at a later time.  In the meanwhile, a cheap improvement would be to do this conditionally if the cloudmaster is the production cloudmaster[1].  I'll do this when merging all minions under a single cloudmaster.

[1] currently we're only tracking the production controller, which is a bit of an anachronism but actually works better in the current situation.  I think I'd rather clean things up than making one right out of two wrongs, though.

salt/flashlight/monitor.bash
```diff
@@ -1,7 +1,7 @@
 #!/bin/bash
 
-# This script sends an email alert when the load average exceeds 0.7. It also
-# logs the load average to statshub.
+# This script sends an email alert when the load average exceeds 70% of the
+# number of cores. It also logs the load average to statshub.
```
aranhoide: Oh, nice!

salt/peerscanner/init.sls
```diff
@@ -0,0 +1,75 @@
+include:
+    - proxy_ufw_rules
+
+{% set zone='getiantem.org' %}
+{% set domain_records_file='/home/lantern/cloudflare_records.yaml' %}
```
aranhoide: Is this needed anymore?

oxtoacart: Ahh, then I guess not.  Let me remove it.

oxtoacart: Removed.

salt/peerscanner/init.sls
```diff
@@ -0,0 +1,75 @@
...
+
+curl:
+  pkg:
+    - installed
+
+mailutils:
+  pkg:
+    - installed
+
+/usr/bin/peerscanner:
```
aranhoide: I think this works only by coincidence.  There is no guarantee that this will be executed before the next state.  I think it is also unnecessary since I've confirmed anecdotically (in Ubuntu 14.10 x64) that curl will overwrite the old file by default.  If you really want to make sure, I'd just prepend a `rm /usr/bin/peerscanner` to the `ps-installed` state.

salt/peerscanner/init.sls
```diff
@@ -0,0 +1,75 @@
+include:
+    - proxy_ufw_rules
```
aranhoide: I don't see any firewall rules added in this PR, so I suppose this is just a C&P artifact.  I'd remove it for clarity (along with a similar one below).

salt/peerscanner/init.sls
```diff
@@ -0,0 +1,75 @@
...
+        - name: 'initctl reload-configuration'
+        - watch:
+            - file: ps-upstart-script
+
+peerscanner:
+    service.running:
+        - enable: yes
+        - require:
+            # All but the last requirement are redundant, only for robustness.
+            - cmd: ufw-rules-ready
```
aranhoide: I don't see any firewall rules added in this PR, so I suppose this is just a C&P artifact.  I'd remove it for clarity (along with a similar one above).

salt/peerscanner/init.sls
oxtoacart: I think there's some weirdness with not being able to replace the file while the service is running.  I wonder if file.absent takes care of that?  How could I make ps-installed dependent on this?

aranhoide: add `- file: /usr/bin/peerscanner` to the `require`s of `ps-installed`.  As I said, a clearer thing would be just to remove it in `ps-installed` before calling `curl`:

` - name: 'rm -f /usr/bin/peerscanner && curl -L https://github.com/getlantern/flashlight-build/releases/download/0.0.9/peerscanner_linux_amd64 -o peerscanner && chmod a+x peerscanner'`

aranhoide:
The only serious problem is the `/usr/bin/peerscanner` being removed independently of it being downloaded again.  It could happen that the state that makes sure that the file doesn't exist is run after the one that downloads it, thus resulting in the executable not being there.

salt/peerscanner/init.sls
oxtoacart: Fixed

salt/peerscanner/init.sls
oxtoacart: Fixed.

salt/peerscanner/init.sls
oxtoacart: I tried just using `rm -f`, but `watch` condition on the `peerscanner` service complains about missing a prerequisite or something, so I went ahead and just added the dependency as you described.  With that, the `watch` condition works and the service restarts as expected.

oxtoacart:
@aranhoide Thanks for the review!  Should be all set now.  I went ahead and also took out some unnecessary jinja templating stuff.

aranhoide:
I'm assuming the second introduction of `- file: /usr/bin/peerscanner` was indeed a typo.  I just removed it, since it was the last think keeping me from merging these.

oxtoacart:
@aranhoide Thanks!  Interesting that it worked in my test, I guess Salt really did just ignore it.

PR #38, opened 2015-03-14
Multifronting

aranhoide:
Changes to salt configuration to support multifronting in flashlight and peerscanners.

Note that this depends on us having released a flashlight and peerscanner with the changes in https://github.com/getlantern/flashlight-build/pull/128.  I'll update the peerscanner and flashlight download URLs when we have released those.


salt/fallback_proxy/upload_wrappers.py
```diff
@@ -1,122 +0,0 @@
-#!/usr/bin/env python
```
aranhoide: This change is unrelated.  An old file that wasn't being used anymore.

salt/peerscanner/peerscanner.conf
```diff
@@ -1 +0,0 @@
-../../secret/peerscanner.conf
```
aranhoide: Only the actual credentials are in the secret repo now.

salt/flashlight/init.sls
```diff
@@ -19,6 +17,10 @@ mailutils:
   pkg:
     - installed
 
+libappindicator3-1:
```
aranhoide: When built directly from the flashlight directory, flashlight requires this.  I don't know whether we'll actually build it that way, but it can't hurt to have this installed just in case.

PR #39, opened 2015-03-27
Save access data

aranhoide:
- Allow us to launch old-style fallback proxies with arbitrary names.
- Make fallback proxies not report to the controller, but save access data locally.
- Add script to the controller that collects access data in all proxies.

I have tested this and I think these are safe changes, so I'll merge it.  I'm submitting this PR just in case anyone wants to give it a look.


README.md
```diff
@@ -110,6 +110,40 @@ To run an arbitrary command (as root) in all fallback proxies:
...
+If all you want to know is whether the machine(s) are done setting themselves
+up (e.g., you haven't made any config changes), you can run something like
+(e.g., for flashlight servers)
+
+    bin/ssh_cloudmaster.py 'salt "fl-20150327-*" cmd.run "service flashlight status"'
+
+If you expect to run a lot of these it will be faster log into the
+cloudmaster (just 'bin/ssh_cloudmaster.py` without arguments) and run the
+commands from there.
+
```
myleshorton: These additions are super helpful @aranhoide thanks!

salt/cloudmaster/access_data.py
```diff
@@ -0,0 +1,32 @@
+#!/usr/bin/env python
```
myleshorton: Can you maybe add a high level comment to this script? Looks like it's handy for getting the access logs is that right? Maybe also add it to the readme? Seems handy =).

aranhoide: Sorry, it's probably badly named.  This is the script you run in the cloudmaster to get the fallbacks.json we used to get from GAE.  That is, all the data clients need in order to access old-style fallbacks (ip, access token, etc.).  I'll certainly add it to the README.

aranhoide: Also, this script has a couple shortcomings but I fixed it in master, so don't pay too much attention to this one.

PR #40, opened 2015-04-09
Vultr

aranhoide:
This branch includes various changes required in order to launch servers out of the typical cloudmaster workflow, and also some Vultr-specific fixes.

This also includes some utilities to launch Vultr machines.

This process is semi-manual and will be superseded by a salt-cloud Vultr module, but some of the changes here will come in handy when we next need to launch servers in a provider that is not yet supported by salt-cloud.


aranhoide:
I'll merge this myself as soon as the first batch of Vultr servers are done installing Salt and we need to apply this configuration to them.  We can only push the production cloudmaster from the master branch.

PR #42, opened 2015-06-12
cron job to check that VPSs in DO and Vultr match what we expect

aranhoide:


aranhoide:
Merging since I've tested this in my test controller, and anyway the only ill effect this seems likely to ever cause is one daily false-alarm email in our fallback-alarms mailing list.

aranhoide:
Feel free to review anyway!

PR #47, opened 2015-08-16
Launchq

aranhoide:
This adds a pair of services that listen for requests from the config server to launch new Lantern Chained Servers in either the Amsterdam (DO) or Tokyo (Vultr) datacenters.  We run one instance of service per datacenter because the end goal (that this PR brings very close, I think) is to have one master per datacenter.

I'll just merge this because I've tested this extensively, but I'm making this PR so you can take a look if you want.  The approach to launching servers used here also obsoletes the Amazon SQS system we were using as a legacy from the time when the lantern-controller used to request new servers to be launched.  I'm leaving the SQS system around for the moment because the redis-based system is only concerned with chained servers (LCSs) and not other types of machines we might want to launch (e.g. fronted flashlight servers, LFSs).


myleshorton:
Very cool @aranhoide thanks. So the config server is the one that queries the existing servers for their load averages and used bandwidth and then splits them accordingly, is that right? And the splitting is completely automated at this point?

aranhoide:
The servers query themselves [1] and ask[2] the config server to split them[3][4][5].  Splitting used to be completely automated already.  What used to be manual, and this PR automates, is refilling the queue of servers ready to be pulled in on split.

As I said, much of this is undocumented, and it should stabilize soon, at which point it will be more effective to do a global writeup of how the system works.  I think a holistic doc would be best; I'll update the READMEs too, but I expect those to be hard to understand separately without an overview of where each part fits.
 
[1] https://github.com/getlantern/lantern_aws/blob/master/salt/fallback_proxy/check_load.py

[2] https://github.com/getlantern/lantern_aws/blob/master/salt/fallback_proxy/util.py#L51

[3] https://github.com/getlantern/config-server/blob/master/src/lantern_config/web.clj#L74

[4] https://github.com/getlantern/config-server/blob/master/src/lantern_config/redis.clj#L103

[5] https://github.com/getlantern/config-server/blob/master/resources/split_ranges.lua

aranhoide:
In addition we have the fallback checker, which would alert us in the case a server crashes so badly that it can't even split itself.  Reacting to output from the fallback checker is manual currently, but this is needed much more rarely.

myleshorton:
Awesome thanks! And where is the code for the Vultr fallbacks checking
their bandwidth?

On Mon, Aug 17, 2015 at 2:08 PM, aranhoide <notifications@github.com> wrote:

> In addition we have the fallback checker, which would alert us in the case
> a server crashes so badly that it can't even split itself. Reacting to
> output from the fallback checker is manual currently, but this is needed
> much more rarely.
>
> —
> <https://github.com/getlantern/lantern_aws/pull/47#issuecomment-131961395>
> .
>



-- 
--
President
https://www.getlantern.org

aranhoide:
It's at https://github.com/getlantern/lantern_aws/blob/master/salt/fallback_proxy/check_vultr_transfer.py

PR #49, opened 2015-08-30
Add collectd to monitor server performance

fffw:
@aranhoide Could you take a look? It's my first attempt to play with salt.

Tested with command `sudo salt '*' state.sls monitor` on an Ubuntu 14.10 VM. Same host acts as master as well as minion.


aranhoide:
Whoops, I must have been on vacation when you submitted this?  I had totally missed it.  I'll get to this as soon as we're done with ATS and CF.

aranhoide:
LGTM, runs without problems after merging master into it.  I suppose you have checked that influx.getlantern.org actually receives these stats.

fffw:
Yes @aranhoide InfluxDB received the stats. I'll have to update InfluxDB to latest version to show over utilized servers through the `TOP()` aggregation, but it's unclear if the `BOTTOM()` aggregation we can use to detect blocked servers is available for now.

PR #50, opened 2015-09-08
Use ATS as chained server

fffw:
It should work if what we want is to replace all servers from lantern-java to ATS. No auth token and cert will be changed when applied to existing servers.


fffw:
@aranhoide mind take a look when you have time?

aranhoide:
I see the reasoning behind making the files depend on the install-ats script (the script creates the directory where the files live) but I still think the directories better be created as a salt state too.

I also realise that Java, which I had removed, must be kept around, because it's needed for keytool at least.  I'll add it back.

Finally, I'll test this config in a machine created from scratch, to make sure we weren't relying on anything else that had been installed in this server, or in a previous run of state.highstate.  Then I'll push my changes.

fffw:
Thanks for the great review @aranhoide !

aranhoide:
Heh, it was mostly useless since you had everything right to begin with!  Although I don't think it's unclear at all to mix both sets of files if install-ats doesn't depend on any of them.

Gotcha on install-ats overriding those files!  I'll keep the order you had (but I'm still thinking of merging both lists of files to avoid repetition; sounds good?)

fffw:
Yes I agree that we can merge both lists. I separated it when `install-ats` depends on auth token but now it is no longer the case.

aranhoide:
@fffw :  in a convoluted way, I merged files, used `creates` instead of `unless: [ -e ... ]` in a couple places, brought Java back, and removed the `lantern` state which installed lantern-java.  I checked that this version installs and proxies fine in a freshly built fallback proxy (fp-nl-fffw-003).  Let me know whether I borked your stuff too badly.  :)

aranhoide:
I've just realised that if we want to deploy this on existing servers we'll need a couple states to at least stop and unregister the lantern-java service.  I'll do that next.

myleshorton:
Awesome @aranhoide. We could also for now just do it for new servers in the queue -- whatever you think is easier.

aranhoide:
I think that would be more of a headache for no gain.  I'm done with the
states that remove old stuff I think.  I'll test it on a lantern-java test
proxy (fp-nl-ara-001).

myleshorton:
Awesome!

On Tue, Sep 8, 2015 at 8:40 AM, aranhoide <notifications@github.com> wrote:

> I think that would be more of a headache for no gain. I'm done with the
> states that remove old stuff I think. I'll test it on a lantern-java test
> proxy (fp-nl-ara-001).
>
> On Tue, Sep 8, 2015 at 5:36 PM, myleshorton <notifications@github.com>
> wrote:
>
> > Awesome @aranhoide <https://github.com/aranhoide>. We could also for now
> > just do it for new servers in the queue -- whatever you think is easier.
> >
> > —
> > Reply to this email directly or view it on GitHub
> > <
> https://github.com/getlantern/lantern_aws/pull/50#issuecomment-138602782>
> > .
>
> >
>
> —
> <https://github.com/getlantern/lantern_aws/pull/50#issuecomment-138603737>
> .
>



-- 
--
President
https://www.getlantern.org

fffw:
@aranhoide `install-ats` will create all those sub directories under `/opt/ts`, with `sudo -u lantern`, so I think the only thing we need for `fp-dirs` target is to chown `/opt/ts` to `lantern`.

fffw:
Sorry I mean make that directory.

aranhoide:
@fffw: I removed the creation of the directories in `install-ats` because I thought it would be best to have it in the Salt config where other stuff can depend on them explicitly.

aranhoide:
OK I think this is done.  I've just sent test YAMLs to the dev list.  Unless you think I borked some of your work, feel free to merge this anytime.

aranhoide:
Great work @fffw !  Impressive for a second shot at Salt.

fffw:
Thanks for the thorough review and additional work @aranhoide ! I'll go ahead and merge it.

PR #51, opened 2015-09-11
more cipher suits to avoid handshake failure

fffw:
Hey @aranhoide do you want to take a look and, probably, run highstate on all chained servers again? It should only affect a few servers but good to do it sometime later.


salt/fallback_proxy/records.config
```diff
@@ -169,7 +169,7 @@ CONFIG proxy.config.reverse_proxy.enabled INT 0
...
+CONFIG proxy.config.ssl.server.cipher_suite STRING HIGH:!DSS:!PSK:!aNULL:!eNULL:!SSLv2
```
fffw: I noticed handshake failure again with fp-nl-20150727-003 (188.166.75.60). Rather than list all possible value, I'd like to follow [NGINX's practice](http://nginx.org/en/docs/http/ngx_http_ssl_module.html#ssl_ciphers), to use all ciphers available except those fatally insecure ones.

salt/fallback_proxy/init.sls
```diff
@@ -180,7 +180,6 @@ ats-service:
     service.running:
         - name: trafficserver
         - enable: yes
-        - reload: yes
```
fffw: Reload doesn't apply some configuration, so I choose to restart service. It will interrupt service for 1-2 seconds, should be ok for users.

salt/fallback_proxy/records.config
aranhoide: Isn't SSLv3 fatally insecure nowadays too?

aranhoide: And, did you update that server manually?  Because it seems to have your new setup already.

aranhoide: @fffw ^

fffw: In fact, I don't have much knowledge on security.

I did see "no shared ciphers" from that server given this long list of ciphers.

Do you think the practice in [this article](https://raymii.org/s/tutorials/Strong_SSL_Security_On_nginx.html) be good enough.

aranhoide: I'm no expert either.  That seems good enough.  It includes advice on disabling SSLv3, by using:

    ssl_protocols TLSv1 TLSv1.1 TLSv1.2;

Are you setting this anywhere?

fffw: Good. Will do it.

aranhoide:
LGTM

PR #52, opened 2015-09-16
add remap rule fixes lantern#3147

fffw:
It's a non-perfect but rather simple solution to fix https://github.com/getlantern/lantern/issues/3147.

Tested on the sandbox salt env.

@aranhoide Would you mind take a look?


myleshorton:
This will still send actual connect data tunneled to the destination, @fffw correct?

fffw:
Yes from what I'm seeing it doesn't affect CONNECT requests.

myleshorton:
Sweet!! Great find.

aranhoide:
Nothing wrong jumps out at me, but I'm not familiar with this ATS stuff, so that's not saying much.

If you've tested it and it fixes the problem, I'm happy to merge it and apply it.  Do you want me to do that, @fffw ?

aranhoide:
Well, if @myleshorton has reviewed the ATS part of this I certainly don't see any problem with the salt part.  Merging and applying...

fffw:
Thanks @aranhoide !

PR #56, opened 2015-09-17
Provide a Redis DB for test deployments of a Salt cloud

uaalto:
A Redis DB is running in Cloudmaster, which is configured to allow connections to the corresponding port from minions. This allows all operations done on Redis, such as minion launching, to happen completely isolated from production.


aranhoide:
Works fine after resolving some merge issues.  It's great to take the production redis URL from non-production cloudmasters.

PR #57, opened 2015-09-20
ATS config updates

fffw:
This will prevent ATS from appending port to "Host" HTTP Header. Fixes https://github.com/getlantern/lantern/issues/3147, again.

As it affects very few sites, it's ok to merge and deploy with other changes.


myleshorton:
LGTM!

PR #58, opened 2015-09-21
mimic ATS default behaviour when auth failed

fffw:
Fixes https://github.com/getlantern/lantern/issues/3133. Can compare the response by `curl`, or through `checkfallbacks` (new PR soon).

The actual code change is at https://github.com/getlantern/ats/pull/1/files in which I already merged.

@myleshorton Do you think it's ok to mimic in this way?


fffw:
Updated binary with changes of https://github.com/getlantern/ats/pull/2

aranhoide:
@fffw : this also includes the removal of the auth header, right?

fffw:
@aranhoide Yes

myleshorton:
Wait sorry @fffw it removes the use of the auth header entirely?

uaalto:
@myleshorton The auth header is removed from the relayed message, so the target server doesn't see it.
Sorry for getting into the conversation, I'm working with this code ATM so, I need to be 100% sure this is what we want! :)

myleshorton:
Oh good -- yeah that make sense thanks @uaalto!

fffw:
Sorry for the confusion and thank you @uaalto for the explanation.

I do consider to move all the ATS config and plugin source code to `lantern_aws` to make each PR more clear and, possibly, avoid leaking too much information to censor. Seems no user is trying to setup their own chained server for now.

uaalto:
I think that's a good idea :) A lot of useful information is here. It just depends if we want to opensource the full stack or not.
In any case, we are moving to the Go version, so ATS is "deprecating".

aranhoide:
Both changes sound good to me.  Since this has had no objections for a week, I'll merge it and deploy it along with a recent PR.

PR #60, opened 2015-10-07
Adjust ATS network parameters

fffw:
@myleshorton @aranhoide This is the changes we discussed in email.

I applied the new configuration to 45.32.249.51 manually. Let's keep it running for a few hours and see if works!

`lsof -nPp 9291 | grep ESTAB | cut -d ">" -f 2 | cut -d ":" -f 1 | sort | uniq -c | sort -nk 1`


myleshorton:
LGTM @fffw. Honestly I think things are so bad in production right now that we should just deploy it.

fffw:
Yeah you are right @myleshorton. We can't see the impact to users by running it on single server. We should deploy it to all servers and see.

myleshorton:
Yup I agree @fffw! I really think it's quite likely to work well. 

I'm also going to investigate the client-side CPU issue today and will keep an eye out for weird connections buildups.

PR #61, opened 2015-10-30
add config for badvpn-udpgw

aranhoide:


salt/fallback_proxy/init.sls
```diff
@@ -30,7 +30,12 @@ fp-dirs:
     ('/opt/ts/etc/trafficserver/', 'records.config', 'records.config', 'lantern', 400),
     ('/opt/ts/etc/trafficserver/', 'remap.config', 'remap.config', 'lantern', 400),
     ('/opt/ts/etc/trafficserver/', 'plugin.config', 'plugin.config', 'lantern', 400),
-    ('/opt/ts/etc/trafficserver/', 'ssl_multicert.config', 'ssl_multicert.config', 'lantern', 400) ]%}
+    ('/opt/ts/etc/trafficserver/', 'ssl_multicert.config', 'ssl_multicert.config', 'lantern', 400) ] %}
+
```
aranhoide: The bunch of files above were (possibly) templatized using Jinja2.  I didn't want that for the new ones I'm adding for udpgw, so I created a new list of verbatim files.

salt/fallback_proxy/udpgw-init
```diff
@@ -0,0 +1,60 @@
...
+test -x $DAEMON || exit 0
+
+PIDFILE=/var/run/$NAME.pid
+
+set -e
+
+case "$1" in
+  start)
+	echo -n "Starting $DESC: "
+	ulimit -n 64000
```
myleshorton: Does this conflict with any other ulimit we're setting anywhere else @aranhoide, @fffw, or @atavism?

salt/fallback_proxy/init.sls
```diff
@@ -30,7 +30,12 @@ fp-dirs:
...
+# To copy verbatim.
+{% set nontemplate_files=[
+    ('/usr/local/bin/', 'badvpn-udpgw', 'badvpn-udpgw', 'root', 755),
```
aranhoide: I compiled the binary in a cloudmaster and, since it was small enough at some 300K, I just added it to the list of files to push.  This way it's version-pinned for reproducibility.

myleshorton: Nice.

salt/fallback_proxy/init.sls
```diff
@@ -55,6 +60,21 @@ include:
             - cmd: install-ats
 {% endfor %}
 
+{% for dir,dst_filename,src_filename,user,mode in nontemplate_files %}
```
aranhoide: This just "unpacks" the config for the files described (more succintly) above, in the format that Salt actually expects.

salt/fallback_proxy/init.sls
```diff
@@ -62,6 +82,9 @@ fallback-proxy-dirs-and-files:
             {% for dir,dst_filename,src_filename,user,mode in template_files %}
             - file: {{ dir+dst_filename }}
             {% endfor %}
+            {% for dir,dst_filename,src_filename,user,mode in nontemplate_files %}
```
aranhoide: Here we add the new files to a dummy state that depends on all files we use and does nothing on its own.  This is just so we can make any state that depends on any file(s) just declare a dependency on this single state instead.

salt/fallback_proxy/init.sls
```diff
@@ -191,7 +214,13 @@ ats-service:
             # it. :)
             - cmd: install-ats
             - service: lantern-disabled
+            - service: badvpn-udpgw
```
aranhoide: We make ATS depend on udpgw so ATS won't start proxying until udpgw is ready.

salt/fallback_proxy/init.sls
```diff
@@ -191,7 +214,13 @@ ats-service:
...
 
+badvpn-udpgw:
+  service.running:
+    - enable: yes
```
aranhoide: This is just the standard way to define services in Salt.  `enable` means that the service should be started up at system boot.

salt/fallback_proxy/udpgw-init
```diff
@@ -0,0 +1,60 @@
...
+	echo -n "Stopping $DESC: "
+	start-stop-daemon -K -q -p $PIDFILE --oknodo
+	echo "$NAME."
+	;;
+  restart)
+	$0 stop
+	sleep 1
+	$0 start
+	;;
+  status)
```
aranhoide: The init file you sent me didn't have a status section.  Salt requires that, and it's generally good to have, so I added it.

salt/fallback_proxy/udpgw-init
fffw: From what described [here](http://ss64.com/bash/ulimit.html), `ulimit` only affects the processes the script spawns, so it would be ok.

salt/fallback_proxy/udpgw-init
atavism: Thanks!

salt/fallback_proxy/init.sls
atavism: That's great. Thanks.

salt/fallback_proxy/udpgw-init
atavism: Yeah, this would only be a per-process limit so it shouldn't affect anything else. We're setting this with a high number of max file descriptors by default as a cautionary measure as too low a limit would typically be much worse.

atavism:
All looks good to me, @aranhoide. Thanks for setting this up!

PR #62, opened 2015-10-31
reduce the data points of collectd

fffw:
It increased the collect interval to 15 minutes, and removed lots of measurements we didn't present for now, to prevent InfluxDB from explosion.


aranhoide:
LGTM.

PR #63, opened 2015-10-31
Retire overloaded servers

aranhoide:
Finding servers with recurring high load average and retiring them takes quite a bit of my everyday time.  I hadn't automated this for a number of reasons; most importantly:

- I wasn't sure what's exactly too high load average.  I actually am not sure where to draw the line yet, but in some cases it seems pretty obvious that servers are overloaded, so there's no point in having to find and retire those servers manually.  For this first attempt, I've chosen a 15m load average of 3.0 or above as a clear indicator that a server should be retired.  Not many servers hit this threshold ever, so even if this was unnecessary, not much harm is done either.

- I worried that automatically retiring fallbacks might cause a domino effect, because the new fallbacks of the dislodged users might get overloaded in turn, resulting in a spiral of fallback retirements.  In a high growth scenario, we could conceivably reach a point where servers were getting retired faster than our refill service could replace them (especially in Vultr, where our API calls are throttled and launching servers often fails at some point and needs to be retried).  I addressed this concern in two ways:

  * an absolute cap: we're not retiring overloaded fallbacks while the `srvq` (the queue of launched but yet unused servers) is too close to empty (currently set as 15 fallbacks or less), and

  * a gap (currently, one day) between splitting and retiring: server load is usually reduced quite a bit after a fallback is split.  This wait prevents us from prematurely retiring servers that wouldn't really be overloaded in the long term.  But the wait also helps dampen any possible domino effect.  Note that dislodged users get reassigned to "open" servers, which would need to get split before they are retired.

Note that these cautionary measures are only for the case when we retire a server because of high load average.  This code doesn't apply for the case when we retire a Vultr server to prevent it from going over quota.  The comments in the code explain why.


aranhoide:
I have tested the code in a fallback managed by my test cloudmaster.  This PR is meant for reviewing that the approach seems sound.

aranhoide:
Since I've tested this and the load average threshold is conservative enough, I'll deploy it.

PR #64, opened 2015-11-10
Go proxy

aranhoide:


salt/fallback_proxy/install_http_proxy.sh
```diff
@@ -0,0 +1,6 @@
+#! /usr/bin/env sh
+
+echo 'Installing golang 1.5.1 ...'
+curl -LO https://github.com/getlantern/http-proxy/releases/download/v0.0.1/http-proxy
```
myleshorton: This doesn't actually install golang right? Is that done somewhere else @aranhoide?

aranhoide: Heh, I saw that printout but forgot to fix it.  That downloads the http-proxy binary.  We aren't installing golang at all currently.

myleshorton: Oh right -- I guess we don't need go with the binary itself! Makes sense @aranhoide.

fffw: Blame me please. I made that confusion.

PR #65, opened 2015-11-13
use http-proxy v0.0.2

fffw:
v0.0.2 includes @uaalto 's fixes to eliminate the infinite loop to localhost. The changes didn't break Apache mimicking, I just tweaked the test code a bit to stop it from hanging, ref PR https://github.com/getlantern/http-proxy-lantern/pull/8.

@myleshorton and @aranhoide, as per our discussion in stand up, it's probably ok to deploy to Vultr?

### DIFF
--- a/salt/fallback_proxy/convcert.sh
+++ b/salt/fallback_proxy/convcert.sh
@@ -6,5 +6,6 @@ KEY_FILE=/home/lantern/key.pem
 CERT_FILE=/home/lantern/cert.pem
 keytool -v -importkeystore -srckeystore $KEYSTORE -srcalias fallback --srcstorepass "$CERT_PASS" -destkeystore /tmp/keystore.p12 -deststoretype PKCS12 --deststorepass "$CERT_PASS"
 openssl pkcs12 -in /tmp/keystore.p12 -passin pass:"$CERT_PASS" -out $KEY_FILE -passout pass:"$CERT_PASS" -nocerts -nodes
-openssl pkcs12 -in /tmp/keystore.p12 -passin pass:"$CERT_PASS" -out $CERT_FILE -passout pass:"$CERT_PASS" -clcerts -nokeys
-rm /tmp/keystore.p12
+openssl pkcs12 -in /tmp/keystore.p12 -passin pass:"$CERT_PASS" -out /tmp/cert.pem -passout pass:"$CERT_PASS" -clcerts -nokeys
+openssl x509 -in /tmp/cert.pem -passin pass:"$CERT_PASS" -out $CERT_FILE -passout pass:"$CERT_PASS"
+rm /tmp/keystore.p12 /tmp/cert.pem
--- a/salt/fallback_proxy/convcert.sh
+++ b/salt/fallback_proxy/convcert.sh
@@ -5,7 +5,6 @@ CERT_PASS="Be Your Own Lantern"
 KEY_FILE=/home/lantern/key.pem
 CERT_FILE=/home/lantern/cert.pem
 keytool -v -importkeystore -srckeystore $KEYSTORE -srcalias fallback --srcstorepass "$CERT_PASS" -destkeystore /tmp/keystore.p12 -deststoretype PKCS12 --deststorepass "$CERT_PASS"
-openssl pkcs12 -in /tmp/keystore.p12 -passin pass:"$CERT_PASS" -out $KEY_FILE -passout pass:"$CERT_PASS" -nocerts -nodes
-openssl pkcs12 -in /tmp/keystore.p12 -passin pass:"$CERT_PASS" -out /tmp/cert.pem -passout pass:"$CERT_PASS" -clcerts -nokeys
-openssl x509 -in /tmp/cert.pem -out $CERT_FILE
-rm /tmp/keystore.p12 /tmp/cert.pem
+openssl pkcs12 -in /tmp/keystore.p12 -passin pass:"$CERT_PASS" -out $KEY_FILE -passout pass:"$CERT_PASS" -nocerts
+keytool -exportcert -alias fallback -storepass "$CERT_PASS" -rfc -keystore /home/lantern/littleproxy_keystore.jks -file $CERT_FILE
+rm /tmp/keystore.p12
--- a/salt/fallback_proxy/convcert.sh
+++ b/salt/fallback_proxy/convcert.sh
@@ -5,6 +5,6 @@ CERT_PASS="Be Your Own Lantern"
 KEY_FILE=/home/lantern/key.pem
 CERT_FILE=/home/lantern/cert.pem
 keytool -v -importkeystore -srckeystore $KEYSTORE -srcalias fallback --srcstorepass "$CERT_PASS" -destkeystore /tmp/keystore.p12 -deststoretype PKCS12 --deststorepass "$CERT_PASS"
-openssl pkcs12 -in /tmp/keystore.p12 -passin pass:"$CERT_PASS" -out $KEY_FILE -passout pass:"$CERT_PASS" -nocerts
+openssl pkcs12 -in /tmp/keystore.p12 -passin pass:"$CERT_PASS" -out $KEY_FILE -passout pass:"$CERT_PASS" -nocerts -nodes
 keytool -exportcert -alias fallback -storepass "$CERT_PASS" -rfc -keystore /home/lantern/littleproxy_keystore.jks -file $CERT_FILE
 rm /tmp/keystore.p12
--- a/salt/fallback_proxy/convcert.sh
+++ b/salt/fallback_proxy/convcert.sh
@@ -7,5 +7,5 @@ CERT_FILE=/home/lantern/cert.pem
 keytool -v -importkeystore -srckeystore $KEYSTORE -srcalias fallback --srcstorepass "$CERT_PASS" -destkeystore /tmp/keystore.p12 -deststoretype PKCS12 --deststorepass "$CERT_PASS"
 openssl pkcs12 -in /tmp/keystore.p12 -passin pass:"$CERT_PASS" -out $KEY_FILE -passout pass:"$CERT_PASS" -nocerts -nodes
 openssl pkcs12 -in /tmp/keystore.p12 -passin pass:"$CERT_PASS" -out /tmp/cert.pem -passout pass:"$CERT_PASS" -clcerts -nokeys
-openssl x509 -in /tmp/cert.pem -passin pass:"$CERT_PASS" -out $CERT_FILE -passout pass:"$CERT_PASS"
+openssl x509 -in /tmp/cert.pem -out $CERT_FILE
 rm /tmp/keystore.p12 /tmp/cert.pem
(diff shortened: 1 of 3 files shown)


salt/fallback_proxy/convcert.sh
```diff
@@ -6,5 +6,6 @@ KEY_FILE=/home/lantern/key.pem
 CERT_FILE=/home/lantern/cert.pem
 keytool -v -importkeystore -srckeystore $KEYSTORE -srcalias fallback --srcstorepass "$CERT_PASS" -destkeystore /tmp/keystore.p12 -deststoretype PKCS12 --deststorepass "$CERT_PASS"
 openssl pkcs12 -in /tmp/keystore.p12 -passin pass:"$CERT_PASS" -out $KEY_FILE -passout pass:"$CERT_PASS" -nocerts -nodes
-openssl pkcs12 -in /tmp/keystore.p12 -passin pass:"$CERT_PASS" -out $CERT_FILE -passout pass:"$CERT_PASS" -clcerts -nokeys
-rm /tmp/keystore.p12
+openssl pkcs12 -in /tmp/keystore.p12 -passin pass:"$CERT_PASS" -out /tmp/cert.pem -passout pass:"$CERT_PASS" -clcerts -nokeys
```
fffw: Oh there was a critical bug which can prevent some client from being able to parse the generated PEM format cert, hence result `No dialers left to try` error. Interestingly the server can parse both the key and cert PEM file.

`openssl pkcs12` will generate a cert with a few extra lines like below:

```
Bag Attributes
    friendlyName: fallback
    localKeyID: 54 69 6D 65 20 31 34 34 37 33 38 34 31 33 30 31 30 34
subject=/C=US/ST=New Jersey/L=Metastasis Atavism/O=Dealings Disorganization Saris/CN=Taper
issuer=/C=US/ST=New Jersey/L=Metastasis Atavism/O=Dealings Disorganization Saris/CN=Taper
```

The `openssl x509` command I added is to remove those lines so Lantern client can parse it.

salt/fallback_proxy/init.sls
```diff
@@ -181,7 +181,7 @@ install-http-proxy:
 convert-cert:
     cmd.script:
         - source: salt://fallback_proxy/convcert.sh
-        - creates: /home/lantern/key.pem
+        - creates: /home/lantern/cert.pem
```
fffw: cert is more likely to fail to generate

fffw:
One pending issue is that `proxy-service` doesn't restart when the key and cert file changes. IIRC the ATS service will restart for any changes. @aranhoide did you experience it before?

```
          ID: proxy-service
    Function: service.running
        Name: http-proxy
      Result: True
     Comment: Service http-proxy is already enabled, and is running
     Started: 03:44:31.954150
    Duration: 106.429 ms
     Changes:
              ----------
              http-proxy:
                  True
```

salt/fallback_proxy/convcert.sh
myleshorton: Wait it looks like this generates `/tmp/cert.pem` while `salt/fallback_proxy/init.sls` below is looking for `/home/lantern/cert.pem`. I don't fully understand how this all works, but it looks like that might generate an error no?

myleshorton: Oh and what was the error with clients not being able to parse the generated PEM format cert @fffw? Very interesting!

myleshorton: Oh sorry now I see now -- the `pkcs12` thing. Wow wait do you have any sense of how widespread that issue was @fffw?

fffw: I should add comment at the next line so it can be shown in conversation.

I found it at the Iran VPS when encountered "No dialers left" error. But now I found it might not be due to the extra lines. The config for `188.166.28.188` messed up and not consistent with `access_data.json` on that server. 

> Nov 13 05:22:11.518 - 0m0s ERROR flashlight.client: balancer.go:48 Unable to configure chained server. Received error: Unable to parse certificate: Unable to decode PEM encoded certificate

myleshorton: Interesting @fffw. I see that error too. There are also all those errors with trying to hit 127 with the server running at 0.0.0.0 -- I can easily fix that one. Would be handy to print out the bad PEM cert there.

OK I'm off to bed, but good and interesting catch!

fffw: Now I confirmed that there's no problem with those extra lines in PEM file. Seems the one got from config server changed `\n` to character `n`.

Below is the full cert, notice the `n`s at the end of each line (others in middle).

```
      cert: 'Bag Attributesn    friendlyName: fallbackn    localKeyID: 54 69 6D 65
        20 31 34 34 37 33 36 36 32 38 33 37 39 39 nsubject=/C=US/ST=Texas/L=Glossing/O=Horrendous
        Hermit/CN=Layer Incorrectlynissuer=/C=US/ST=Texas/L=Glossing/O=Horrendous
        Hermit/CN=Layer Incorrectlyn-----BEGIN CERTIFICATE-----nMIIB9zCCAZygAwIBAgIEJ30DjjAMBggqhkjOPQQDAgUAMGgxCzAJBgNVBAYTAlVTn
        MQ4wDAYDVQQIEwVUZXhhczERMA8GA1UEBxMIR2xvc3NpbmcxGjAYBgNVBAoTEUhvncnJlbmRvdXMgSGVybWl0MRowGAYDVQQDExFMYXllciBJbmNvcnJlY3RseTAeFw0xn
        NTA5MjkyMjExMjNaFw0xNzA5MjgyMjExMjNaMGgxCzAJBgNVBAYTAlVTMQ4wDAYDnVQQIEwVUZXhhczERMA8GA1UEBxMIR2xvc3NpbmcxGjAYBgNVBAoTEUhvcnJlbmRvn
        dXMgSGVybWl0MRowGAYDVQQDExFMYXllciBJbmNvcnJlY3RseTBZMBMGByqGSM49nAgEGCCqGSM49AwEHA0IABIeb5O21MqHpNHDpW1Cgew/JQ3lAgY1IXrxuFPhk4Pxan
        fQfdmLgDgs8oPQ7C4PFrmHalzpMqNZITmA9frG8JAgSjMjAwMA8GA1UdEQQIMAaHnBLymHLwwHQYDVR0OBBYEFKWkAeD4GHGh/S4k4JcOp/+IilhuMAwGCCqGSM49BAMCn
        BQADRwAwRAIgdfK9ZNTekK4Mpm7w5Mc0ftt3SBRyp+oMNzaR45IeJA4CIHdHNY3pnuudKn/8Bc2q9uAAo554Tfd7RdsO1TGUUSNW1n
        -----END CERTIFICATE-----n'
```

myleshorton: Very interesting @fffw. That sounds like it could cause global problems depending on how the client deals with it. It's weird there are extra "n"s on all those lines: `Bag Attributesn`, `fallbackn`, `Incorrectlyn`, etc. I'm guessing there's some issue with one of the scripts.

fffw: Ok I'll check a few servers to find out the reason. Good sleep myleshorton!

salt/fallback_proxy/convcert.sh
```diff
@@ -5,6 +5,6 @@ CERT_PASS="Be Your Own Lantern"
 KEY_FILE=/home/lantern/key.pem
 CERT_FILE=/home/lantern/cert.pem
 keytool -v -importkeystore -srckeystore $KEYSTORE -srcalias fallback --srcstorepass "$CERT_PASS" -destkeystore /tmp/keystore.p12 -deststoretype PKCS12 --deststorepass "$CERT_PASS"
-openssl pkcs12 -in /tmp/keystore.p12 -passin pass:"$CERT_PASS" -out $KEY_FILE -passout pass:"$CERT_PASS" -nocerts -nodes
-openssl pkcs12 -in /tmp/keystore.p12 -passin pass:"$CERT_PASS" -out $CERT_FILE -passout pass:"$CERT_PASS" -clcerts -nokeys
+openssl pkcs12 -in /tmp/keystore.p12 -passin pass:"$CERT_PASS" -out $KEY_FILE -passout pass:"$CERT_PASS" -nocerts
```
fffw: I can't remember when I added the `-nodes` option, but it means `don't encrypt private keys` which is not what we want.

salt/fallback_proxy/convcert.sh
```diff
@@ -5,6 +5,6 @@ CERT_PASS="Be Your Own Lantern"
...
+keytool -exportcert -alias fallback -storepass "$CERT_PASS" -rfc -keystore /home/lantern/littleproxy_keystore.jks -file $CERT_FILE
```
fffw: This is exactly what `save_access_data.py` did previously, so the `pkcs12` specific `Bag Attributes` will not appear in cert file, and eventually, the cert return from config server.

salt/fallback_proxy/convcert.sh
fffw: I compared the `access_data.json` file of 188.166.28.188 (the one assigned to Iran VPS which failed to decode PEM) and 45.32.249.211 (an normal JP server for which `ip2yaml.py` gets correct data), no much difference. Both should work. The character `n` is a mystery.

salt/fallback_proxy/convcert.sh
myleshorton: OK so does this all work now @fffw?

aranhoide: Wouldn't checkfallbacks.py have caught any problems with certs or any other part of access data?

Of course that'd be only for newly launched servers.  Servers upgraded from ATS already have their (old) access data in redis.

aranhoide: Which reminds me, I failed to test this in a newly launched server; I only checked that upgrading an ATS server worked fine, but that wouldn't catch any errors exporting the cert to access_data.json, because that's only done on server creation, not (by default) reconfiguration.

aranhoide:
@fffw : right now the http-proxy service should restart *every time* a salt config is applied, just because the binary is downloaded every time, and the service state is watching that.  When I fix that, I'll look at the other dependencies.

aranhoide:
LGTM; I'm testing it on both an upgraded server and a newly launched one.

aranhoide:
@fffw : one problem with this config, I think, is that if a key.pem exists, it will never be regenerated again even if the convcert script or the original cert changes.  Not a big deal since we can always remove those manually.

salt/fallback_proxy/convcert.sh
aranhoide: It seems http-proxy can't load the private key if it's encrypted, so we probably want `-nodes` after all (unless we fix this in http-proxy).

I'm not sure why we should encrypt this key anyway?  The secrets to decrypt that key will need to live in that box anyway, so I think if the box is compromised we're out of luck already?

aranhoide:
OK, a newly created fallback works fine with these changes.  Merging and deploying to Amsterdam...

myleshorton:
Awesome @aranhoide!

salt/fallback_proxy/convcert.sh
fffw: Sorry I thought I tested the generated key, but apparently I didn't. Thanks
for catching it!

fffw:
What I saw from salt-call log is, http-proxy binary downloaded each time,
but the old process is still there. Seems the process didn't get killed
during service restarting, though 'kill <pid>' manually could work.

PR #66, opened 2015-11-18
Allow several data centers per 'user region'

aranhoide:


aranhoide:
I've tested refilling the server queue, retiring and destroying servers, so I'm merging this.

myleshorton:
Sweet

PR #67, opened 2015-11-19
Download http-proxy-lantern v0.0.3 from private repo

fffw:
Please pull secrets repo and `git submodule update --remote` first.

I tried testing it on sandbox cloudmaster + minion, but it always failed to install packages.

> WARNING: The following packages cannot be authenticated!
>   unzip zip
> There are problems and -y was used without --force-yes

### DIFF
--- a/bin/update.py
+++ b/bin/update.py
@@ -91,6 +91,7 @@ def upload_pillars():
     _, _, do_token = util.read_do_credential()
     vultr_apikey = util.read_vultr_credential()
     cfgsrv_token, cfgsrv_redis_url = util.read_cfgsrv_credential()
+    github_token = util.read_github_token()
     if not util.in_production():
         cfgsrv_redis_url = "redis://%s:6379" % config.cloudmaster_address
     util.ssh_cloudmaster((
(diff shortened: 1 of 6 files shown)


bin/update.py
```diff
@@ -106,8 +107,9 @@ def upload_pillars():
             ' && echo "vultr_apikey: %s" > vultr_credential.sls'
             ' && echo "cfgsrv_token: %s" > cfgsrv_credential.sls'
             ' && echo "cfgsrv_redis_url: %s" >> cfgsrv_credential.sls'
-           r' && echo "base: {\"*\": [salt, global], \"fp-*\": [cfgsrv_credential, vultr_credential], \"cm-*\": [do_credential, vultr_credential, cfgsrv_credential]}" > top.sls '
-            ' && sudo mv salt.sls global.sls top.sls do_credential.sls vultr_credential.sls cfgsrv_credential.sls $(hostname).sls /srv/pillar/ '
+            ' && echo "github_token: %s" > github_token.sls'
+            r' && echo "base: {\"*\": [salt, global], \"fp-*\": [cfgsrv_credential, vultr_credential], \"cm-*\": [do_credential, vultr_credential, cfgsrv_credential, github_token]}" > top.sls '
```
fffw: Oh the `github_token` pillar should be added only to `fp-*` instead right?

aranhoide: Whoops, you're right.  I missed that one too.

An alternative would be that only the cloudmaster downloads the binary from github as part of its own setup, and then all the fallbacks get http-proxy from it as a regular salt file.

fffw: That would be good, though it will add 10M transfer for each minion.

fffw: It's not clear to me how to do it. During `bin/update.py`?

aranhoide: To begin with, we can leave this as a refinement for later, because it still involves the `install_http_proxy.py` script.  Only that would run in the cloudmaster, and it would download the file to `/srv/salt/http_proxy/` instead.  Then in `lantern_aws/salt/http_proxy/init.sls` we could add it as a regular file with `source: salt://http_proxy/http-proxy`.

aranhoide: So, in order to update all proxies, we would first need to run highstate in the cloudmaster, then in the proxies.

fffw:
I made a few changes and now it can download http-proxy binary from test minion.

bin/update.py
aranhoide: An advantage of doing it this way is that proxies would only download this when there is an updated version.  But it's good enough as it is currently, so maybe let's make a note to improve this later.

aranhoide:
Nice!  Yeah, I missed the jinja part too.

I don't get the errors you get about installing zip, etc.  Maybe that's because my test cloudmaster is in Amsterdam?

aranhoide:
I don't know how you're testing this exactly, but for this to apply to fallbacks you need to substitute `http-proxy` for `ats` in `lantern_aws/salt/top.sls`.  Were you doing this locally?

aranhoide:
This works flawlessly for me in Amsterdam.  I'll now test in Tokyo, to see whether I get any issues with apt-get.

fffw:
Yes I changed `top.sls` manually.

aranhoide:
Vultr works perfectly too for me.  No apt-get errors with zip or otherwise.  Your test fallback must be in a weird state?

Anyway, great work!  Merging and deploying to Amsterdam.

fffw:
Great thanks a lot @aranhoide

PR #68, opened 2015-11-20
fetch http-proxy by tag

fffw:
I didn't notice the existence of this specific GitHub API endpoint. It's simpler than I thought.


fffw:
Tested on a sandbox minion

aranhoide:
This is great, thanks!  I double-verified that it works.  I'll merge, but no point in deploying it to existing fallbacks since the exact same binary would be downloaded.

fffw:
Sure. Not worth deploying. Thanks Alan!

PR #69, opened 2015-11-22
supply loggly token to http-proxy

fffw:
http-proxy already has `-logglytoken` option, this PR supply our own token when running it.

### DIFF
--- a/salt/http_proxy/init.sls
+++ b/salt/http_proxy/init.sls
@@ -1,6 +1,7 @@
 {% set fallback_json_file='/home/lantern/fallback.json' %}
 {% set proxy_protocol=pillar.get('proxy_protocol', 'tcp') %}
 {% set auth_token=pillar.get('auth_token') %}
+{% set loggly_token=pillar.get('loggly_token') %}
 {% set proxy_port=grains.get('proxy_port', 62443) %}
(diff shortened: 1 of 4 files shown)


salt/http_proxy/init.sls
```diff
@@ -1,6 +1,7 @@
 {% set fallback_json_file='/home/lantern/fallback.json' %}
 {% set proxy_protocol=pillar.get('proxy_protocol', 'tcp') %}
 {% set auth_token=pillar.get('auth_token') %}
+{% set loggly_token=pillar.get('loggly_token') %}
```
aranhoide: We could have read the pillar in the file itself.  I added these variables originally to avoid duplication (although maybe some of them are not necessary anymore).

aranhoide:
LGTM, I have checked that it works, and taken the liberty of removing the `loggly_token` variable, and reading that pillar from the templated file instead.

fffw:
Thanks @aranhoide !

aranhoide:
Also, sorry that adding pillars is so cumbersome!  It started as a quick hack and it didn't occur to me that we'd be touching that so often.  I'll clean it up a bit next time we add one. :)

PR #72, opened 2015-12-16
stop http-proxy before updating its binary

aranhoide:
Occasionally, we get `text file busy` errors when trying to start the http-proxy service.  I've noticed this once after upgrading the http-proxy-lantern release.  I suspect there's something dangerous about updating the binary of a service while it's running.  I remember we used to stop the old (flashlight, peerscanner, ...) services before updating them, for this reason.

### DIFF
--- a/salt/http_proxy/install_http_proxy.py
+++ b/salt/http_proxy/install_http_proxy.py
@@ -25,8 +25,13 @@ if(r.ok):
             for chunk in r.iter_content(chunk_size=1048576):
                 if chunk:
                     f.write(chunk)
+        os.chmod('http-proxy-temp', stat.S_IRUSR | stat.S_IWUSR | stat.S_IXUSR)
+        # We have noticed some errors, which may have to do with the binary
+        # being updated while the service is running.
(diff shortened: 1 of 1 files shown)


aranhoide:
Could you review this when you're done with Pro tasks, @fffw?

This is not super high priority, since I can check for this condition whenever I deploy an update to fallbacks.

aranhoide:
@fffw: I've found a different solution which is preferrable if it works; I'm testing it now, so please don't review this just yet...

aranhoide:
OK, unlinking the old executable before renaming the new one into it seems to work, and it's less intrusive than using `sudo` to stop the service (which may give us a bit more downtime).  @fffw this is ready for review whenever you have the time (again, this is not really urgent).

aranhoide:
BTW I got that solution from

http://stackoverflow.com/questions/1712033/replacing-a-running-executable-in-linux

fffw:
Yeah it's a classical unix trick! LGTM. Thanks for sorting it out @aranhoide !

aranhoide:
Thanks!

PR #79, opened 2016-02-18
Country specific req stats

aranhoide:
Updates to the caching code used by retire_blocked_ips, to support country-specific request stats.

Also sneaked in: retry some 10 times when Vultr server_list(None) returns the empty list.


aranhoide:
I'll go ahead and merge this.  I've already deployed the config-server part so the old version of this logic is useless.

PR #81, opened 2016-02-22
quarantine blocked IPs

aranhoide:
Store VPSs in blocked IPs in a quarantine set, instead of using them to serve users.  Flush the quarantine into the destroy queue when we don't have any pending or ongoing requests.


aranhoide:
As I said in today's checkin, I'm deploying this.

PR #82, opened 2016-02-23
close overloaded proxies instead of splitting them

aranhoide:


salt/http_proxy/check_load.py
```diff
@@ -16,7 +16,7 @@
 
 retire_threshold = 2.0
 report_threshold = 1.6
-split_threshold = 1.5
+close_threshold = 1.2
```
aranhoide: Conservative threshold while we test this.

aranhoide:
I've just tested this in a test cloudmaster/proxy and it works fine.

PR #84, opened 2016-02-26
save last week's load_avg/CPU/mem/network/... stats once a minute

aranhoide:
It's useful to have these on the VPS itself for manual inspection.  Curnently our Influx DB records data less often, keeps less of a history of it, and VPSs are more likely to be unable to hit influx for any reason than to run this cron job.

### DIFF
--- a/salt/base_prereqs.sls
+++ b/salt/base_prereqs.sls
@@ -20,6 +20,7 @@ base-packages:
             - git
             - mailutils
             - debconf-utils
+            - python-psutil
         - reload_modules: yes
 
 /usr/local/lib/pylib:
--- a/salt/http_proxy/init.sls
+++ b/salt/http_proxy/init.sls
@@ -106,8 +106,6 @@ requests:
 
 {% if pillar['in_production'] %}
 
-python-psutil:
-    pkg.installed
 
 uptime:
     pip.installed
@@ -121,7 +119,6 @@ uptime:
         - pip: requests
         - pip: uptime
         - pkg: python-redis
-        - pkg: python-psutil
 
 "/home/lantern/check_traffic.py 2>&1 | logger -t check_traffic":
   cron.absent:
@@ -131,7 +128,6 @@ uptime:
 #    - user: lantern
 #    - require:
 #        - file: /home/lantern/check_traffic.py
-#        - pip: psutil
 
 {% if pillar['datacenter'].startswith('vl') %}
 
--- a/salt/stats/init.sls
+++ b/salt/stats/init.sls
@@ -1,0 +1,11 @@
+/usr/bin/save_stats.py:
+  file.managed:
+    - source: salt://stats/save_stats.py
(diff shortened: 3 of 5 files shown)


aranhoide:
I tested this in a test cloudmaster.  Deploying to all machines.

myleshorton:
Where does it save it @aranhoide? It seems like it would be useful to a
documentation page for analyzing single servers that would include this
like:

1) the path to this file
2) a link to the grafana page where we can select this server
3) the instructions for fetching the auth-token for a given server and
running Lantern to hit it
4) notes on things like syslog or http-proxy log errors we've seen causing
issues in the past
5) more I'm not thinking of!

I can get that started. Ulysses based on your past work on improving docs
where do you think this should live?

uaalto:
I think that if the changes affect what is described [here](https://github.com/getlantern/lantern/blob/valencia/README-dev.md) we should update it first. That's the central documentation for everything. Then a link to https://github.com/getlantern/lantern_aws/blob/master/README.md with a section for this topic, or if it's a big topic, we can make a link in the main README to README-server-analysis or similar.

aranhoide:
@myleshorton: it saves it locally to `/home/lantern/stats`.  This is done for all VPSs managed by Salt, not only proxies.  I don't know how to deep-link to a grafana page for a particular proxy.  Do you know if it's possible, @fffw?  AFAIK you need to go to the general one and then type its name in a search box.

@uaalto: Agreed on the AWS name.  I'll change that after things have settled with blocking.

fffw:
@aranhoide You are right our Grafana page doesn't have link for single
proxy. We can do that by programming on Grafana API if required.

aranhoide:
@fffw: that would help a lot. :)

aranhoide:
Related to this, a new utility function to get a single-figure digest of the traffic in the latest x minutes.

https://github.com/getlantern/lantern_aws/commit/4502dd9d5b1682e93ddcadacb035d14db9dadaad#commitcomment-16369709

fffw:
@aranhoide So I come up with simple dynamic dashboard. 
For example, visit https://dashboard.getlantern.org/dashboard/script/single-proxy.js?name=fp-jp-20150915-028 you'll see a simple dashboard for that specific server.

To align with this PR, I extended the retention policy of proxy metrics from 24 hours to 7 days, but we probably will hit performance bottleneck again, though I just upgraded InfluxDB.

aranhoide:
I don't think we need to keep both records in sync.  The proxies save their stats every minute, which would probably be prohibitive to run in a centralized server.

I think it's good to have a little redundancy here, because this information is so critical.  Influx is great to feed into grafana and get visualizations.  On-proxy storage allows us to keep more data per machine and has different failure cases.

myleshorton:
This is fantastic @fffw! Very cool.

aranhoide:
Very cool indeed! :)

PR #94, opened 2016-03-06
replace email alerts with slack + logging to redis

aranhoide:
This also removes the ats module, which was lying around (and contained references to sending email).


aranhoide:
Tested in test cloudmaster and proxy, so merging.

PR #96, opened 2016-03-09
Issue85

aranhoide:
Also redirect stderr to stdin, so all go to syslog.  Just in case, limit the size of /var/log/upstart/http-proxy.log, which is where stderr used to go.

I tested this vs a script that spammed syslog in a test cloudmaster.  A caveat: logs will not be trimmed when rotated, so we'll keep the whole logs of the latest 5h no matter what.

### DIFF
--- a/salt/logrotate/logrotate.cron
+++ b/salt/logrotate/logrotate.cron
@@ -1,0 +1,14 @@
+#!/bin/sh
+
+# Clean non existent log file entries from status file
+cd /var/lib/logrotate
+test -e status || touch status
+head -1 status > status.clean
+sed 's/"//g' status | while read logfile date
+do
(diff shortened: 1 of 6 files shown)


salt/logrotate/logrotate.cron
```diff
@@ -0,0 +1,14 @@
+#!/bin/sh
```
aranhoide: This is just the standard script that comes with Ubuntu 14.04, unmodified.  This is the same for DO and Vultr.  I'm adding it here because it seems more robust than moving it around in the minion filesystem.

oxtoacart:
LGTM!

myleshorton:
Nice @aranhoide!!

PR #99, opened 2016-03-10
Multislice proxies

aranhoide:
Changes to the lantern_aws project to support assigning multiple slices to one proxy.


bin/check_lcs_queues.py
```diff
@@ -1,139 +0,0 @@
-"""Check LCServer queues in the config server, vs the list of live servers."""
```
aranhoide: Obsolete utilities.

aranhoide:
Merging and deploying to test backwards compatibility prior to actually introducing slices with the new convention.

PR #110, opened 2016-03-14
move 'etc' to donyc3, assign its current cloudmasters to 'ir'

aranhoide:
The transition plan is:

 - shut down all services in existing cloudmasters
 - back up the redis DB
 - run a script to transition the redis DB
 - deploy this to all cloudmasters (this restarts services with the new config)


aranhoide:
Backed up redis and shut down services, about to deploy this after redis transition job...

PR #116, opened 2016-03-23
Issue112

### LINKED ISSUE #112: Make per-cloudmaster launch request queue
...in addition to the per-user-region queues.

While we're at it, we can make server launch requests optionally include size and prefix to override the defaults.  Launching anything but a default-sized proxy is unnecessarily cumbersome now.

aranhoide:
Have per-cloudmaster server request queues and server queues, fixes #112.

Also make the staging condition more visible across the deployment, mostly to make it more like a production setup.  Staging gets more services enabled than a plain test setup, and gets a dedicated slack channel for alerts.

### DIFF
--- a/bin/update.py
+++ b/bin/update.py
@@ -94,12 +94,15 @@ def upload_pillars():
         = util.read_cfgsrv_credential()
     github_token = util.read_github_token()
     loggly_token = util.read_loggly_token()
-    slack_webhook_url = util.read_slack_webhook_url()
+    if util.in_production():
+        slack_webhook_url = util.read_slack_webhook_url()
+    else:
+        slack_webhook_url = util.read_slack_staging_webhook_url()
     if not util.in_production():
-        if config.cloudmaster_name == "cm-dosgp1staging":
-            # Exception: make the staging cloudmaster in Singapore use the
-            # redis instance of the staging cloudmaster in Amsterdam, to be
-            # more like the production setup.
+        if util.in_staging():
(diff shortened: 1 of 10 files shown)


bin/update.py
```diff
@@ -94,12 +94,15 @@ def upload_pillars():
         = util.read_cfgsrv_credential()
     github_token = util.read_github_token()
     loggly_token = util.read_loggly_token()
-    slack_webhook_url = util.read_slack_webhook_url()
+    if util.in_production():
+        slack_webhook_url = util.read_slack_webhook_url()
+    else:
+        slack_webhook_url = util.read_slack_staging_webhook_url()
```
aranhoide: I could make a `testing` webhook for deployments not in staging or production, but we basically ~never get a log from such deployments, so this would only clutter the Slack channels list.

aranhoide:
Tested in staging, and I'm pretty sure this is backwards-compatible, so merging, deploying, and double-checking in production.

myleshorton:
Exciting!

On Wed, Mar 23, 2016 at 6:40 PM, aranhoide <notifications@github.com> wrote:

> Tested in staging, and I'm pretty sure this is backwards-compatible, so
> merging, deploying, and double-checking in production.
>
> —
> You are receiving this because you are subscribed to this thread.
> <https://github.com/getlantern/lantern_aws/pull/116#issuecomment-200608696>
>



-- 
--
President
https://www.getlantern.org

PR #122, opened 2016-03-28
Added obfs4 support (don't merge yet)

oxtoacart:
I've moved these changes into branch + PR like I should have


aranhoide:
Thanks!  Then again, if you had tested the changes it's OK; as I said in the comment before, I was just double-checking because the naming of the http-proxy release looked like it might not be intended for production deployment.

aranhoide:
LGTM.  I have tested the old path by launching a proxy with no obfs4 pillars, so if you have tested the obfs4 version I think this is good to go.

PR #123, opened 2016-03-29
[don't merge] Publish IPv6 address(es) in config

aranhoide:
Depends on https://github.com/getlantern/http-proxy-lantern/issues/48 and https://github.com/getlantern/lantern_aws/issues/124.  **Don't merge** before those are resolved.

Note that as of this writing all new proxies are created with IPv6 enabled.  That was already true for Vultr, and I've made it so earlier today for DO.  This PR is only about feeding that info to the config that proxies publish for the clients' benefit.

Currently only one address is added, but we could conceivably add more in the future.  In any case, the client may forward-compatibly choose to use them all, only the first one, or none.

I've tested this in both DO and Vultr proxies, both newly created and updated, both with and without IPv6 support.  (To be precise, I've tested the version that only publishes one address, but that particular change is trivial and vendor-independent.)

### DIFF
--- a/lib/alert.py
+++ b/lib/alert.py
@@ -6,8 +6,8 @@ import requests
 
 
 instance_id = "{{ grains['id'] }}"
-# {% from 'ip.sls' import external_ip %}
-ip = "{{ external_ip(grains) }}"
+# {% from 'ip.sls' import external_ipv4 %}
+ip = "{{ external_ipv4(grains) }}"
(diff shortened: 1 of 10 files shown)


aranhoide:
Inheriting priority from overarching issue https://github.com/getlantern/lantern/issues/3775

aranhoide:
Closing this one because it's rotting here and is only distracting us.

PR #125, opened 2016-03-30
add netdata support; enable in all VPSs

aranhoide:
Tested in my test cloudmaster.

### DIFF
--- a/salt/netdata/install-netdata.bash
+++ b/salt/netdata/install-netdata.bash
@@ -1,0 +1,13 @@
+#!/bin/bash
+
+set -e
+
+DIR=$(mktemp -d)
+
+cd $DIR
+git clone --depth=1 git://github.com/firehol/netdata.git
+cd netdata
+./netdata-installer.sh --dont-start-it --dont-wait --zlib-is-really-here --install /opt
+cd /tmp
+rm -rf $DIR
+touch /etc/netdata-installed
--- a/salt/netdata/install-netdata.bash
+++ b/salt/netdata/install-netdata.bash
@@ -5,7 +5,7 @@ set -e
 DIR=$(mktemp -d)
 
 cd $DIR
-git clone --depth=1 git://github.com/firehol/netdata.git
+git clone --branch=v1.0.0 --depth=1 git://github.com/firehol/netdata.git
 cd netdata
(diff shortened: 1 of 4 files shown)


salt/netdata/install-netdata.bash
```diff
@@ -0,0 +1,13 @@
+#!/bin/bash
+
+set -e
+
+DIR=$(mktemp -d)
+
+cd $DIR
+git clone --depth=1 git://github.com/firehol/netdata.git
```
oxtoacart: Do we want to trust that master is always good, or grab a particular tag/commit hash?  Per the netdata instructions, it seems master is good at least for now, so this is probably fine.

salt/netdata/netdata.conf
```diff
@@ -0,0 +1,27 @@
...
+start on runlevel [2345]
+
+# When to stop the service
+stop on runlevel [016]
+
+# Automatically restart process if crashed
+respawn
+
+# Terminate with SIGINT instead of SIGKILL
+kill signal SIGINT
```
oxtoacart: I don't know how netdata responds to this.  This is something custom we did for our proxy process so that its SIGINT handler could do its thing.  Probably not necessary for netdata.

aranhoide: Good point; this was just c+p from the http_proxy config; I'll remove it and test whether shutdown works well.

salt/netdata/install-netdata.bash
aranhoide: Yes, good point.  There seems to be a 1.0.0 release from 8 days ago, so I'll use that instead for reproducibility.

aranhoide:
@oxtoacart I fixed the issues you pointed out.  I also noticed that netdata was forking into the background, so while it worked, it was not really managed by upstart anymore.  I fixed that too.

oxtoacart:
LGTM!

aranhoide:
OK, I'll deploy this to all cloudmasters and proxies.

@oxtoacart if your pubsub is salt-managed you may want to clean its current installation of netdata before ever calling highstate on it again.

Then again, I don't expect anything terrible will happen if you don't.  The salt/upstart-managed version will fail to grab the port and it will give up after a few tries.  Then it will take over on next reboot (netdata doesn't have any autostart setup; that's the main reason I put it on upstart).

PR #126, opened 2016-03-31
upgrade to java 8

aranhoide:
The only thing for which we're currently using Java is for generating certs in proxies.  I have launched a new proxy from this branch and it passed the checkfallbacks test.

The main impetus for this is that Java 8 makes building ops-panel easier.  I'm getting [permgen errors](https://github.com/boot-clj/boot/wiki/JVM-Options#permgen-errors-java-7) while doing this.  While there are workarounds for that, it seems simplest to just upgrade to a Java version without this problem.

### DIFF
--- a/salt/lantern_build_prereqs/init.sls
+++ b/salt/lantern_build_prereqs/init.sls
@@ -5,26 +5,26 @@ java-ppa:
 java-home:
     file.append:
         - name: /etc/profile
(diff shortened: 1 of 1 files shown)


aranhoide:
I'll merge and deploy this since it's a very mechanical change and I've tested the only part it affects.

PR #127, opened 2016-03-31
Added support for launching obfs4 proxies from srvreqq (incompatible with current configserver)

oxtoacart:
This requires a new request format.  Instead of requesting just a number, this now takes a JSON data structure with the following parameters:

```
{
    "id": 6,
    "obfs4_port": 62443,
    "srvq": "donyc3ox:srvq:manual",
}
```

id - this is the same as the number that we used to send.
obfs4_port - optional port for launching an obfs4 proxy instead of an https proxy.  This should always be 62443.
srvq - the queue to which the successful launch response is sent.  This used to always go to the same queue, but now client can specify it.

### DIFF
--- a/lib/vps_util.py
+++ b/lib/vps_util.py
@@ -209,7 +209,9 @@ def cm_by_name(name):
     elif name.startswith('fp-jp-'):
         name = name.replace('jp', 'vltok1', 1)
 
-    return name.split('-')[1]
+    # We need to count from the right because we have HTTPS proxies both with
+    # and without the -https- part in their name.
+    return name.split('-')[-3]
 
 def region_by_name(name):
     return region_by_dc(dc_by_name(name))
--- a/lib/vps_util.py
+++ b/lib/vps_util.py
@@ -22,6 +22,7 @@ auth_token: %s
 install-from: git
 instance_id: %s
 proxy_protocol: tcp
+obfs4_port: %s
 """
 
 auth_token_alphabet = string.ascii_letters + string.digits
@@ -44,9 +45,10 @@ def random_auth_token():
     return ''.join(random.choice(auth_token_alphabet)
                    for _ in xrange(auth_token_length))
 
-def save_pillar(name):
+def save_pillar(name, req):
+    obfs4_port = req.get('obfs4_port', 0)
     with file("/srv/pillar/%s.sls" % name, 'w') as f:
-        f.write(pillar_tmpl % (random_auth_token(), name))
+        f.write(pillar_tmpl % (random_auth_token(), name, obfs4_port))
 
 def trycmd(cmd, tries=sys.maxsize):
     for x in xrange(tries):
@@ -226,7 +228,8 @@ _region_by_production_cm = {'donyc3': 'etc',
                             'vlpar1': 'ir',
                             'dosgp1': 'sea',
                             'dosfo1': 'sea',
-                            'vltok1': 'sea'}
+                            'vltok1': 'sea',
+                            'vllan1': 'etc',}
 def region_by_dc(dc):
     return _region_by_production_cm[dc]
 
--- a/salt/cloudmaster/refill_srvq.py
+++ b/salt/cloudmaster/refill_srvq.py
@@ -8,6 +8,7 @@ import time
 import traceback
 
 import yaml
+import json
 
 from alert import send_to_slack
 from redis_util import redis_shell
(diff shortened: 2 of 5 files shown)


lib/vps_util.py
```diff
@@ -226,7 +228,8 @@ def dc_by_cm(cm):
                             'vlpar1': 'ir',
                             'dosgp1': 'sea',
                             'dosfo1': 'sea',
-                            'vltok1': 'sea'}
+                            'vltok1': 'sea',
+                            'vllan1': 'etc',}
```
aranhoide: Whoops, missed this one, sorry!

oxtoacart:
I tested this on Digital Ocean and Vultr by submitting a launch request from REDIS, like this:

```
RPUSH vllan1ox:srvreqq '{ "id": 3, "obfs4_port": 62443, "srvq": "vllan1ox:srvq:manual" }'
```

I tested with an HTTPS and an OBFS4 proxy.  The OBFS4 proxy ran into an issue with the http-proxy binary, but I'm resolving that separately.  Otherwise, everything seemed fine.

salt/cloudmaster/refill_srvq.py
```diff
@@ -83,17 +84,20 @@ def kill_task(reqid):
                 print "Wat?"
                 break
         if len(pending) < MAXPROCS:
-            reqid, remover = reqq.next_job()
-            if reqid:
-                print "Got request", reqid
+            req_string, remover = reqq.next_job()
+            if req_string:
+                print "Got request", req_string
+                req = json.loads(req_string)
```
aranhoide: In the old format, reqid would parse to a number.  Transition would be easier and safer if we supported this in an intermediary stage.  This should work:
```
if isinstance(req, int):
   req = {'id': req, 'srvq': QPREFIX + ':srvq'}
```

salt/cloudmaster/refill_srvq.py
```diff
@@ -121,14 +125,17 @@ def get_lcs_name():
         pipe.set(CM + ':lcsserial', 1)
         pipe.execute()
         serial = 1
-    return 'fp-%s-%s-%03d' % (CM, date, serial)
+    type_prefix = 'obfs4' if 'obfs4_port' in req else 'https'
+    return 'fp-%s-%s-%s-%03d' % (type_prefix, CM, date, serial)
```
aranhoide: This change breaks `vps_util.dc_by_name` (and some Pro server code).  I'm not sure what's the best solution here.  Thinking about it...

secret
```diff
@@ -1 +1 @@
-Subproject commit e31ff8b245723d3ed69e5c0ec794f18e70884cc9
+Subproject commit 3aa7d884d3b5ac4107c59d7bc26d77f4a06f8999
```
aranhoide: What secrets are added here?

aranhoide:
@oxtoacart I'm done with the review.  The naming convention issue is thorny.  `fp-` is currently assumed in a lot of places, so removing the `-` there doesn't seem safe.

One thing we could do is to identify the cloudmaster/datacenter by index from the right.  Hopefully all current proxies will have been replaced before we need to break that too.  :P

lib/vps_util.py
```diff
@@ -209,7 +209,9 @@ def cm_by_name(name):
     elif name.startswith('fp-jp-'):
         name = name.replace('jp', 'vltok1', 1)
 
-    return name.split('-')[1]
+    # We need to count from the right because we have HTTPS proxies both with
+    # and without the -https- part in their name.
+    return name.split('-')[-3]
```
aranhoide: Whoops, I meant to make this a sub-PR to merge into `obfs4launch` (or into master, doesn't matter a whole lot).

Anyway, please let me know whether this looks good to you.  If so, I'll submit an equivalent PR to the Pro server (which, being in Clojure, can't share this code).

secret
oxtoacart: Whatever's on master of too-many-secrets.

aranhoide: OK, works for me.  I was just double-checking, should I have missed some new secret.

lib/vps_util.py
oxtoacart: Yes, this looks good.

secret
myleshorton: Hey @oxtoacart does it always use 62443? Seems like that would make it easy
to block.

oxtoacart: 62443 is internal. External is 443, which seems to be recommended for obfs4.

-- 
Cheers,
Ox

PGP Key - https://pgp.mit.edu/pks/lookup?op=get&search=ox@getlantern.org

-----------------------------------------------------------------------------------------

"I love people who harness themselves, an ox to a heavy cart,
who pull like water buffalo, with massive patience,
who strain in the mud and the muck to move things forward,
who do what has to be done, again and again."

- Marge Piercy

PR #128, opened 2016-03-31
avoid breaking config server et al while in transition

aranhoide:
Ugly, but should be short-lived, since the change to config server et al should be quick.


oxtoacart:
I like it!

PR #132, opened 2016-04-06
join cert to single line fix getlantern/lantern#3915

fffw:


salt/http_proxy/save_access_data.py
```diff
@@ -44,7 +46,7 @@ def add_obfs4_access_data(d_out):
             if m:
                 d_out['cert'] = m.group(1)
```
fffw: The cert of `obfs4` should be in single line as they are matched using `regex`.

aranhoide:
LGTM!  Have you tested launching a new server with this config?

fffw:
@aranhoide No I'm not that familiar with the process to launch a new server. I verified only by executing that Python snippet directly.

aranhoide:
OK, I'm trying it; I'll document the process to launch a new server tomorrow morning.

aranhoide:
@fffw: I created a new test proxy (188.166.38.120) with this configuration, and for some reason it's failing fallback checks.

I'm exhausted now; I'll look into this tomorrow morning.

aranhoide:
A variation of this that just normalizes on single '\n's passes checkfallbacks and the uploaded cert looks right, so I'll push this.

fffw:
Cool thanks so much @aranhoide

PR #135, opened 2016-04-06
utilities to fix access_data/cfg issues; related refactoring

aranhoide:
This was created to fix some cert issues.  It's probably wise to keep these around, so if we have any similar problems in the future they're faster to fix.

The only tricky part about this is that some refactoring was done to share some logic with refill_srvq.py.  I have checked that launching proxies works alright with this new factoring.

This is not urgent, but preventing branch rot would be nice.

### DIFF
--- a/lib/vps_util.py
+++ b/lib/vps_util.py
@@ -390,4 +390,4 @@ def fix_live_servers():
             fixed_access_data = fix_cert_newlines(access_data)
             fixed_cfg = serialize_access_data(fixed_access_data, name)
             fixed[srv] = fixed_cfg
-    redis_shell.hmget('srv->cfg', fixed)
+    redis_shell.hmset('srv->cfg', fixed)
--- a/lib/vps_util.py
+++ b/lib/vps_util.py
@@ -323,3 +323,71 @@ def assign_clientip_to_new_own_srv(clientip, region):
 
 def is_production_proxy(name):
     return name.startswith('fp-') and cm_by_name(name) in _region_by_production_cm
+
+
+def serialize_access_data(access_data, name):
+    ad = access_data.copy()
+    # DRY: flashlight/genconfig/cloud.yaml.tmpl
+    ad.update(pipeline=True,
+              trusted=True,
+              qos=10,
+              weight=1000000)
+    return "\n    " + yaml.dump({name: ad})
+
+def enqueue_cfg(name, access_data, srvq):
+    "Upload a config to a server queue."
+    ip = access_data['addr'].split(':')[0]
+    cfg = serialize_access_data(access_data, name)
+    txn = redis_shell.pipeline()
+    txn.hset('server->config', name, cfg)
+    txn.lpush(srvq, "%s|%s|%s" % (ip, name, cfg))
+    txn.execute()
+
+def fix_cert_newlines(d):
+    d = d.copy()
+    d['cert'] = re.sub(r'[\r\n\\]+', '\n', d['cert'])
+    return d
+
+def fix_access_data(qentry, fixfn):
+    ip, name, cfg = qentry.split('|')
+    access_data = yaml.load(cfg).values()[0]
+    fixed_access_data = fixfn(access_data)
+    return '|'.join([ip, name, serialize_access_data(fixed_access_data, name)])
+
+def filter_queue(qname, fn):
+    bakname = qname + '.bak'
+    redis_shell.rename(qname, bakname)
+    p = redis_shell.pipeline()
+    for entry in redis_shell.lrange(bakname, 0, -1):
+        p.lpush(qname, fn(entry))
+    p.execute()
+
+def fix_queues():
+    regions = redis_shell.smembers('user-regions')
+    cloudmasters = redis_shell.smembers('cloudmasters')
+    for domain in regions | cloudmasters:
+        qname = domain + ':srvq'
+        if redis_shell.exists(qname):
+            print "fixing queue for %s..." % domain
+            fix_queue(qname)
+        else:
+            print "no queue for %s." % domain
+
+
+# Ad hoc application of the building blocks above. Leaving them around, should
+# they serve as inspiration for future fixing jobs.
+
+def fix_queue(qname):
+    filter_queue(qname, lambda s: fix_access_data(s, fix_cert_newlines))
+
+def fix_live_servers():
+    srv2cfg = redis_shell.hgetall('srv->cfg')
+    fixed = {}
+    for srv, cfg in srv2cfg.iteritems():
+        if '\\' in cfg:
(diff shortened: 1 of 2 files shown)


aranhoide:
@oxtoacart as I say above this is not urgent.  Assigning to you because @fffw already stayed awake until late yesterday to debug the related issue, and all other devs are chickens :).  Reviewing the new functionality is not necessary since it'll probably need to be adapted next time it'll be used anyway; only backwards compatibility needs an eye (and I've tested that).

lib/vps_util.py
```diff
@@ -323,3 +323,71 @@ def assign_clientip_to_new_own_srv(clientip, region):
...
+    return "\n    " + yaml.dump({name: ad})
+
+def enqueue_cfg(name, access_data, srvq):
+    "Upload a config to a server queue."
+    ip = access_data['addr'].split(':')[0]
+    cfg = serialize_access_data(access_data, name)
+    txn = redis_shell.pipeline()
+    txn.hset('server->config', name, cfg)
+    txn.lpush(srvq, "%s|%s|%s" % (ip, name, cfg))
+    txn.execute()
```
aranhoide: Only the above is used routinely.  The code below is there to draw from (possibly with some adaptation) next time we need to backport config fixes to existing live or enqueued proxies.

lib/vps_util.py
```diff
@@ -323,3 +323,71 @@ def assign_clientip_to_new_own_srv(clientip, region):
...
+    fixed_access_data = fixfn(access_data)
+    return '|'.join([ip, name, serialize_access_data(fixed_access_data, name)])
+
+def filter_queue(qname, fn):
+    bakname = qname + '.bak'
+    redis_shell.rename(qname, bakname)
+    p = redis_shell.pipeline()
+    for entry in redis_shell.lrange(bakname, 0, -1):
+        p.lpush(qname, fn(entry))
+    p.execute()
```
aranhoide: The backup queue is only deleted manually.  This is on purpose.

lib/vps_util.py
```diff
@@ -323,3 +323,71 @@ def assign_clientip_to_new_own_srv(clientip, region):
 
 def is_production_proxy(name):
     return name.startswith('fp-') and cm_by_name(name) in _region_by_production_cm
+
+
+def serialize_access_data(access_data, name):
+    ad = access_data.copy()
```
oxtoacart: Since we're only updating top-level properties, this shallow copy seems okay to me.  You too @aranhoide ?

lib/vps_util.py
```diff
@@ -323,3 +323,71 @@ def assign_clientip_to_new_own_srv(clientip, region):
...
+    ip, name, cfg = qentry.split('|')
+    access_data = yaml.load(cfg).values()[0]
+    fixed_access_data = fixfn(access_data)
+    return '|'.join([ip, name, serialize_access_data(fixed_access_data, name)])
+
+def filter_queue(qname, fn):
+    bakname = qname + '.bak'
+    redis_shell.rename(qname, bakname)
+    p = redis_shell.pipeline()
+    for entry in redis_shell.lrange(bakname, 0, -1):
```
oxtoacart: I assume that lrange here only iterates over what was originally in the list and doesn't include what we lpush down below?

lib/vps_util.py
```diff
@@ -323,3 +323,71 @@ def assign_clientip_to_new_own_srv(clientip, region):
...
+    srv2cfg = redis_shell.hgetall('srv->cfg')
+    fixed = {}
+    for srv, cfg in srv2cfg.iteritems():
+        if '\\' in cfg:
+            name, access_data = yaml.load(cfg).items()[0]
+            print "fixing srv %s..." % name
+            fixed_access_data = fix_cert_newlines(access_data)
+            fixed_cfg = serialize_access_data(fixed_access_data, name)
+            fixed[srv] = fixed_cfg
+    redis_shell.hmget('srv->cfg', fixed)
```
oxtoacart: Should this be hmset?

oxtoacart:
@aranhoide I finished my first pass, please see my comments.  Also note that there are merge conflicts at the moment.

lib/vps_util.py
aranhoide: Yes; even copying at all seems on the cautious side.

lib/vps_util.py
aranhoide: Yes, lrange eagerly fetches the requested items and returns a list.

lib/vps_util.py
aranhoide: Yes, it should! :D  It's fixed in a later commit.

lib/vps_util.py
```diff
@@ -390,4 +390,4 @@ def fix_live_servers():
             fixed_access_data = fix_cert_newlines(access_data)
             fixed_cfg = serialize_access_data(fixed_access_data, name)
             fixed[srv] = fixed_cfg
-    redis_shell.hmget('srv->cfg', fixed)
+    redis_shell.hmset('srv->cfg', fixed)
```
aranhoide: @oxtoacart this is the fix for the hmget bug.

lib/vps_util.py
oxtoacart: Yeah, you actually fixed this before I found it - somehow I was looking at outdated code.

oxtoacart:
@aranhoide Other than the merge conflicts, this LGTM.

lib/vps_util.py
```diff
@@ -391,3 +390,22 @@ def fix_live_servers():
...
+    while True:
+        entry = redis_shell.rpop(srvq)
+        ip, name, cfg = entry.split('|')
+        print "popping", name
+        next_dc = dc_by_name(name)
+        if next_dc == dc:
+            print "pushing back", name
+            redis_shell.rpush(srvq, entry)
+            return
+        retire_proxy(name=name, ip=ip, reason="Flushing server queue")
```
aranhoide: @oxtoacart : this function at the end of the master version of vps_util was causing the conflict.  Fortunately, its functionality is related to this branch, so I just pulled it in.

Again, this was already run (to good effect, AFAICT).  The only issue with this is that a proxy may be left in limbo if we crash between the time we pull it and the time we push it back or retire it.  Since this is run manually, we can fix this manually.  Also, automated checks and/or compaction jobs will take care of these.

aranhoide:
@oxtoacart : fixing the merge conflicts involved importing yet another utility function from master.  I pushed it there mistakenly assuming I was in this branch.  Could you take a quick look at that too?

aranhoide:
Whoops, I didn't see you'd merged it.  Oh well, this function was in master already after all, and again, chances are I'll change it anyway next time I need to do something similar.

aranhoide:
@oxtoacart Thanks for the review!

PR #136, opened 2016-04-13
add whitelist to SSH alerts, utility SSH wrapper

aranhoide:
Adds a whitelist so cloudmasters and admins don't get alerts when logging in via SSH.

Cloudmasters are in this whitelist permanently.  If someone has root access to our cloudmasters, we are in big trouble already.

`bin/ssh_cloudmaster.py` adds a temporary entry to this whitelist.

For manual operation, `bin/hussh` wraps the regular `ssh` to populate this whitelist with a short expiration (it doesn't matter how long your session is, since the check for both log in and log out is performed at the beginning).


myleshorton:
Hey @uaalto maybe you could take a look at this one?

lib/histate.py
```diff
@@ -1,150 +0,0 @@
-# Note this requires python3!
```
aranhoide: This module was unused.  The only reason it's removed in this PR is that I had forgotten about it.  It used to do SSH, so it came to my attention when looking for things to update.

lib/misc_util.py
```diff
@@ -48,21 +48,56 @@ def __getattr__(self, name):
     def __setattr__(self, name, value):
         return self.__setitem__(name, value)
 
-def ssh(ip, cmd, timeout=60):
+def ssh(ip, cmd, timeout=60, whitelist=True):
+    if whitelist:
+        whitelist_ssh()
+    return cmd_error_and_output(
```
aranhoide: This changes the signature of the function and breaks some functions in secret_lib, but those are only ever called manually, so I'll update them next time I need them, if ever.

salt/sshalert/init.sls
```diff
@@ -6,6 +6,8 @@
 /etc/ssh/sshd_config:
   file.append:
     - text: "ForceCommand /usr/bin/sshalert.py"
+    # prevent spurious logs on VPS creation.
+    - order: last
```
aranhoide: Without this, a single SSH alert would be logged on proxy creation.

aranhoide:
I forgot to mention that I've tested both `ssh_cloudmaster.py` and `hussh` in a test cloudmaster, and tested that launching a proxy doesn't generate SSH login alerts.

uaalto:
Sure I'll take a look @myleshorton !

salt/sshalert/sshalert.py
```diff
@@ -4,18 +4,27 @@
 import sys
 
 import alert
+try:
+    from redis_util import redis_shell
+except ImportError:
+    # sshalert is enabled on all vpss, some of which don't have redis_shell.
```
uaalto: Which don't have redis_shell and why?

aranhoide:
I've just found a bug with `hussh`.  It will work fine for executing a single command, but not for interactive use: in that case, you won't get to see the shell prompt or output (although your commands will be executed).  I'll fix this.

salt/sshalert/sshalert.py
aranhoide: Any machine which is not a cloudmaster or proxy will typically not have access to redis.  The only example currently is pubsub instances.  This is to limit privileges to where they're needed.  As an alternative, we could have different redis instances.  Currently that seems overkill, since I'm not expecting a whole lot of alert noise from logging into pubsub instances.

lib/misc_util.py
```diff
@@ -48,21 +48,56 @@ def __getattr__(self, name):
...
+                     timeout))
+
+def whitelist_ssh(time=60):
     try:
-        return subprocess.check_output(['timeout', str(timeout), 'ssh', '-o', 'StrictHostKeyChecking=no', 'lantern@' + ip, cmd])
+        from redis_util import redis_shell as redis_shell
+    except ImportError:
+        print >> sys.stderr, "No redis access; I won't be whitelisted in this SSH session"
+        return
+    ip = subprocess.check_output(['dig', '+short', 'myip.opendns.com', '@resolver1.opendns.com']).strip()
```
uaalto: Is this the way we learn from our own public ip? If we are not using private networks it should be easy to get also from the interface. Or is it inconsistent across providers?

lib/misc_util.py
```diff
@@ -48,21 +48,56 @@ def __getattr__(self, name):
...
+        try:
+            # This is unlikely to change in Python 2. All interface breaking
+            # changes are happening in 3. If this does change the impact will
+            # be minor (either some log noise or some whitelist entries lasting
+            # longer).
+            poolsize = max(1, int(pool._processes))
+        except AttributeError:
+            pass
+    if whitelist:
+        timeout = 60 + (5 * 60 * len(ips) // poolsize)
```
uaalto: Why the timeout increases proportionally to the number of concurrent ips, and only when we use the whitelist?

lib/misc_util.py
aranhoide: Currently this code is only ever used from our own personal computers.  That's also why it's being looked up per request, since our laptop IPs may change.  Cloudmaster IPs are permanently whitelisted, so cloudmasters don't use this.

If this changes in the future this should still work, although it'd be suboptimal.

One case in which this would stop working would be if we started to use this code in communication between our VPSs *and* we started to use private networks in our datacenters.  Then again, this would only cause ssh alert noise, so we can worry about it when it actually happens.

lib/misc_util.py
```diff
@@ -48,21 +48,56 @@ def __getattr__(self, name):
...
+
+def whitelist_ssh(time=60):
     try:
-        return subprocess.check_output(['timeout', str(timeout), 'ssh', '-o', 'StrictHostKeyChecking=no', 'lantern@' + ip, cmd])
+        from redis_util import redis_shell as redis_shell
+    except ImportError:
+        print >> sys.stderr, "No redis access; I won't be whitelisted in this SSH session"
+        return
+    ip = subprocess.check_output(['dig', '+short', 'myip.opendns.com', '@resolver1.opendns.com']).strip()
+    redis_shell.setex('sshalert-whitelist:%s' % ip, 'admin', time)
```
aranhoide: This argument order is the one redis-py uses.  It differs from the order in redis's own SETEX.

lib/misc_util.py
aranhoide: This is actually a brain fart of mine :).  I didn't really mean to shadow the passed in `timeout` with this one.  I meant to use this one for the expiration of the whitelist for the whole pssh operation.  The intention was to set the whitelist entry only once instead of per ssh call.

I'll fix this; let's see if it makes more sense then.

lib/misc_util.py
```diff
@@ -48,21 +48,60 @@ def __getattr__(self, name):
...
+            # longer).
+            poolsize = max(1, int(pool._processes))
+        except AttributeError:
+            pass
+    if whitelist:
+        # Optimization: we do a single whitelist call instead of len(ips) ones.
+        #
+        # We set a conservative expiration for this one. The whole operation
+        # should have concluded or timed out by this time.
+        whitelist_ssh(time=60 + (timeout * len(ips) // poolsize))
```
aranhoide: @uaalto: this is what I really meant to do here.

salt/sshalert/sshalert.py
uaalto: I agree on limiting privileges.

lib/misc_util.py
uaalto: Sounds good.

lib/misc_util.py
uaalto: :) great!

uaalto:
I'll wait for your fix of `hussh` for its review and merge.

aranhoide:
@uaalto the fix to `hussh` is in.

uaalto:
LGTM. Merging.

aranhoide:
@uaalto thanks for reviewing!  I'll deploy this now.

PR #139, opened 2016-04-19
add requirements.txt resolve #138

### LINKED ISSUE #138: virtualenv support?
It's not uncommon to me that python scripts fails because of lacking certain package, but the error message is not always that apparent. One example: `bin/hussh` fails because package `transit` required by `lib/redis_util.py` is not installed, but I can only see `No redis access; I won't be whitelisted in this SSH session`.

`virtualenv` + `requirements.txt` seems a good fit here. No longer need to install packages in global environment, and any new dependencies can be solved by `pip install -r requirements.txt`.

If no opposition @aranhoide I can make that simple change.

fffw:
@aranhoide So here's the simple PR

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,5 +1,5 @@
 python-dateutil
 python-digitalocean
 redis
-transit-python
-vultr
+transit-python==0.8.282
+vultr==0.1.2
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,0 +1,5 @@
+python-dateutil
+python-digitalocean
+redis
+transit-python
+vultr
--- a/.gitignore
+++ b/.gitignore
@@ -12,3 +12,5 @@ hslog
 bin/result
 bin/status
 .#*
+venv
+.venv
--- a/README.md
+++ b/README.md
@@ -23,9 +23,15 @@ If the fallback proxy fails to set itself up, eventually the SQS message that tr
 
 ### Before you start
 
-You need a recent version of the Digital Ocean API Python wrapper to run the script that launches cloud masters.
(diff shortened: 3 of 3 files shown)


requirements.txt
```diff
@@ -0,0 +1,5 @@
+python-dateutil
+python-digitalocean
+redis
+transit-python
+vultr
```
aranhoide: Have you tested this?  Esp. in a system without python-vultr?

We need to pin the version of at least Vultr.  Our current usage is incompatible with their master branch.  AFAICT no release has made it to pip with that version, but as soon as that happens, that will break things for whoever installs Vultr thereafter.  Transit also has some nonstandard recommendations (`--use-wheel --pre`) on how it should be installed.  I've just tried uninstalling and installing without those arguments, and I seem to get the same version (`transit-python==0.8.282`) without problems.  So maybe that's OK.

Simply replacing `vultr` with `vultr==0.1.2` should fix the Vultr issue.  I would pin Transit to the version above just in case.  We're deprecating Transit anyway since there's no good library for golang AFAIK.

fffw: I previously set version for each package but removed when I saw `pip install -U` in documents. Sure if certain package needs to be locked we can do it.

aranhoide:
@fffw : I've taken the liberty to pin the versions of vultr and transit, for the reasons explained in the commit comment.  I've tested installing the requirements with these changes.

Other than that, this LGTM, so unless you have any objections to my change feel free to merge.

fffw:
Thanks @aranhoide !

aranhoide:
Thanks to you!

PR #146, opened 2016-04-22
ssh-alert doublecheck

aranhoide:
If you accidentally ssh into a proxy, you'll be prompted to use hussh instead.  You can still override this if you're in a rush to fix something or you can't set up hussh at this moment.

For interactive logins, you just get asked to confirm that you really want to log in.  For noninteractive logins, you can prepend `_force_` to bypass this check.  In either case, the script will try and shame you on that. :)

As a side feature, ssh alerts now go to a new #ssh-alert slack channel.

I've tested this in a test proxy.

### DIFF
--- a/salt/sshalert/sshalert.py
+++ b/salt/sshalert/sshalert.py
@@ -24,7 +24,6 @@ if not whitelisted:
     # test cloudmaster).  You can easily lock yourself out of the test machine.
     # Keeping a backup of this file and a separate SSH session open while you
     # test this should help prevent such lockout.
-    force = False
     if not original_cmd:
         print >> sys.stderr, "Your IP is not whitelisted.  Please use lantern_aws/bin/hussh instead."
         print >> sys.stderr, "If you continue, ssh-alert noise will ensue.  You'll need to warn the"
--- a/salt/sshalert/sshalert.py
+++ b/salt/sshalert/sshalert.py
@@ -4,6 +4,7 @@ import os
 import sys
 
 import alert
+import misc_util
 try:
     from redis_util import redis_shell
 except ImportError:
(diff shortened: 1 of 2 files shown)


aranhoide:
This is not very urgent; I only did it now because I thought it would be a way quicker change.

salt/sshalert/sshalert.py
```diff
@@ -15,16 +16,48 @@
...
+original_cmd = os.getenv('SSH_ORIGINAL_COMMAND')
+
 if not whitelisted:
-    alert.send_to_slack("SSH login", "User %s *logging in* from IP %s" % (user, ip))
+    # WARNING: if you make any changes to this check or its response you better
+    # test them in a more disposable machine (e.g., a test proxy, rather than a
+    # test cloudmaster).  You can easily lock yourself out of the test machine.
+    # Keeping a backup of this file and a separate SSH session open while you
+    # test this should help prevent such lockout.
+    force = False
```
aranhoide: I've just realized that this default is not necessary since `force` is set in every branch below.

salt/sshalert/sshalert.py
```diff
@@ -15,16 +16,47 @@
...
+original_cmd = os.getenv('SSH_ORIGINAL_COMMAND')
+
 if not whitelisted:
-    alert.send_to_slack("SSH login", "User %s *logging in* from IP %s" % (user, ip))
+    # WARNING: if you make any changes to this check or its response you better
+    # test them in a more disposable machine (e.g., a test proxy, rather than a
+    # test cloudmaster).  You can easily lock yourself out of the test machine.
+    # Keeping a backup of this file and a separate SSH session open while you
+    # test this should help prevent such lockout.
+    if not original_cmd:
```
uaalto: we are testing whether this is SSH or HuSSH by looking at an environment variable, which I would assume is set by HuSSH, but couldn't find it. How is it set?

salt/sshalert/sshalert.py
```diff
@@ -15,16 +16,47 @@
...
+            print >> sys.stderr, '\n   ssh <my-address> "_force_ %s"\n' % original_cmd
+            force = False
+    if force:
+        print >> sys.stderr, "Please leave a note about this login in #ssh-alerts (preferrably)"
+        print >> sys.stderr, "or in the dev mailing list."
+    else:
+        sys.exit(1)
+    alert.send_to_slack("SSH login",
+                        "User %s *logging in* from IP %s" % (user, ip),
+                        channel="#ssh-alerts")
```
uaalto: Maybe we can avoid this in non-production machines. Is it easily done?

uaalto:
Additionally to the comments I did, I think we should document this, otherwise might become an obscure feature that nobody (except you, if you use it often) remembers about.

uaalto:
One more thing: I've tried logging in with SSH, and it does understand that you aren't using HuSSH properly. However, my intrusion gets logged in `#ssh-alerts`. I expect this to happen because the check is performed in the test cloudmaster local Redis. I want to double-check: is that the case?

salt/sshalert/sshalert.py
aranhoide: We are testing in redis whether the user's IP is whitelisted.  Env vars are used to get the user's IP (and, less to the point, to get the command they are trying to run).  These are not provided by us but by SSH etc.  I'm on my phone but you can google their names.

aranhoide: TBC, we're not testing for hussh specifically.  hussh whitelists your IP in redis temporarily and then calls regular ssh.

salt/sshalert/sshalert.py
aranhoide: The point is to be able to test this in non-production machines.  Also, we don't want random people logging into those either.

Of course logs shouldn't go to the same channel as in production, so I need to test that.

aranhoide:
There's a PR to document the need to use hussh.  I'll link to it when I'm back home.  This feature just makes that harder to ignore.  I think it's self-documenting.  It's for people who missed the docs or forgot about ssh whitelisting.

salt/sshalert/sshalert.py
uaalto: Ok, noted on the SSH_ORIGINAL_COMMAND front. Didn't know this variable was set by SSH. Seems fine to me.

uaalto:
Ok that's enough then. I agree that _force_ is self-documenting since it is displayed to the user directly. Even better than documentation ;)

Just waiting for your reply on my last comment regarding the Slack channel and I'll be ready to merge.

aranhoide:
If you log in without your IP being whitelisted (be it via hussh or other means) then your intrusion will be logged.  That's what ssh-alert is for.

hussh isn't cuŕently smart about test redis instances, if that's what you're asking.  The simplest solution would be to permanently whitelist ourselves in the cloudmasters we're using.

uaalto:
Yes, that was my question. I think is fine, at least for now, given the usage of test cloudmasters currently and that this messages are harmless anyway.

uaalto:
LGTM. Merging.

uaalto:
For the record, I tried this on the test cloud.

aranhoide:
Thanks @uaalto!

PR #147, opened 2016-04-25
remove cloudmaster's key from root's authorized keys as soon as the V…

aranhoide:
…PS is set up

To make this possible, some code has been changed not to use that key any more.

I'm convinced this vulnerability is currently highly theoretical.  To obtain this key, an attacker probably needs to have gained root access to a cloudmaster or login access to our Vultr or DO admin page, both of which are about as dangerous as this key.

It's still wise to remove it, lest it becomes more of a vulnerability in the future.

I've tested this in my test cloudmaster.

### DIFF
--- a/lib/do_util.py
+++ b/lib/do_util.py
@@ -12,7 +12,6 @@ import vps_util
 
 
 ssh_tmpl = "ssh -o StrictHostKeyChecking=no -i /etc/salt/cloudmaster.id_rsa root@%s '%%s'"
-fetchaccessdata_tmpl = "scp -o StrictHostKeyChecking=no -i /etc/salt/cloudmaster.id_rsa root@%s:/home/lantern/access_data.json ."
 
 do_token = os.getenv("DO_TOKEN")
 do = digitalocean.Manager(token=do_token)
@@ -37,10 +36,7 @@ def init_vps(d):
         print("Highstate still running...")
         time.sleep(10)
     print("Highstate done!")
-    return vps_util.hammer_the_damn_thing_until_it_proxies(
-        name,
-        ssh_tmpl % ip,
-        fetchaccessdata_tmpl % ip)
+    return vps_util.hammer_the_damn_thing_until_it_proxies(name)
 
 def droplets_by_name():
     return {d.name: d
--- a/lib/do_util.py
+++ b/lib/do_util.py
@@ -11,8 +11,6 @@ import misc_util
 import vps_util
 
 
-ssh_tmpl = "ssh -o StrictHostKeyChecking=no -i /etc/salt/cloudmaster.id_rsa root@%s '%%s'"
-
 do_token = os.getenv("DO_TOKEN")
 do = digitalocean.Manager(token=do_token)
 
--- a/lib/vps_util.py
+++ b/lib/vps_util.py
@@ -342,7 +342,8 @@ def serialize_access_data(access_data, name):
               trusted=True,
               qos=10,
               weight=1000000)
-    return "\n    " + yaml.dump({name: ad})
+    # Use safe_dump to avoid `!!python/unicode` markers for unicode strings.
+    return "\n    " + yaml.safe_dump({name: ad})
 
 def enqueue_cfg(name, access_data, srvq):
     "Upload a config to a server queue."
--- a/lib/vps_util.py
+++ b/lib/vps_util.py
@@ -1,6 +1,7 @@
 from contextlib import contextmanager
 from datetime import datetime
 import itertools as it
+import json
 import os
 import random
 import re
@@ -9,6 +10,7 @@ import string
 import sys
 import tempfile
 import time
+
 import yaml
 
 import redis_util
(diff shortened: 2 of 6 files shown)


salt/security.sls
```diff
@@ -39,3 +39,10 @@ reload-sshd-on-password-disable:
     - watch:
         - file: disable-password-authentication
         - file: disable-challenge-response-authentication
+
+/root/.ssh/authorized_keys:
```
oxtoacart: What is this doing?  Might warrant a comment, since it's non-obvious.

aranhoide: This removes the `cloudmaster` public key from the root's authorized keys.  That key is identified by having the email lanterncyborg@gmail.com.

I could delete authorized_keys altogether, but if we ever want to accept any SSH key for the root user and we forget about this, it could get really frustrating to debug.

I'll add a comment to this effect when I'm back at my computer.

aranhoide: The rationale is that by the time we have successfully installed Salt and set up this minion with its cloudmaster, the ssh key has performed its function and we don't need it anymore.  The rest of this PR is about making that assumption stand.

lib/do_util.py
```diff
@@ -12,7 +12,6 @@
 
 
 ssh_tmpl = "ssh -o StrictHostKeyChecking=no -i /etc/salt/cloudmaster.id_rsa root@%s '%%s'"
```
oxtoacart: This still references root, but ssh_tmpl doesn't appear to be used anywhere so you could probably just get rid of it.

oxtoacart: Oops, strike that, it is still used.

oxtoacart: Actually, double reversal - there's an ssh_tmpl function that's used, but the ssh_tmpl string appears to be unused.

oxtoacart:
@aranhoide I assume you tested this.  Other than adding a comment about the authorized_keys removal and maybe getting rid of the ssh_tmpl string, this LGTM.

lib/do_util.py
aranhoide: Yes, the cloudmaster key is used for initializing the VPSs, and only once Salt is set up and connected to the master is the key deleted.

In `vultr_util` we need to refer to the key in code because we initialize the VPS directly using the Vultr API.  For DO this initialization is performed by salt-cloud, so while the key is [still used](https://github.com/getlantern/lantern_aws/blob/master/salt/salt_cloud/cloud.providers#L6) we don't need to mention it here.

I'll delete this line, thanks!

aranhoide:
@oxtoacart : I'd tested this up until server creation, which involves checkfallbacks check.  Alas, this leaves one last step untested, and I've just realized there's a problem there.  There are some spurious `!!python/unicode` YAML extension markers.  I'll fix this.

lib/vps_util.py
```diff
@@ -67,36 +69,46 @@ def highstate_pid(name):
...
-                file('fallbacks.json', 'w').write("[" + access_data + "]")
-                for tries in xrange(3):
-                    out = subprocess.check_output(['checkfallbacks',
-                                                   '-fallbacks', 'fallbacks.json',
-                                                   '-connections', '1'])
-                    if "[failed fallback check]" in out:
-                        print("Fallback check failed; retrying...")
-                    else:
-                        print("VPS up!")
-                        return yaml.load(access_data)
```
aranhoide: Actually, I've just remembered that there was a good reason why I was parsing this with the YAML library and not the JSON one.  Turns out that JSON returns all strings as unicode objects, which produce `!!python/unicode` markers when encoding to YAML.  I'll revert this particular change and add a comment to that effect.

aranhoide: Actually, I'll do a more proper fix, using `yaml.safe_dump` when reencoding the file.

aranhoide:
@oxtoacart: I've added the comment you suggest and got rid of ssh_tmpl.  I've also fixed a bug (`!!python/unicode` markers creeping into the config uploaded to redis) which I had missed because it happens after the checkfallbacks check.  I've tested this final change (and this time around I checked the config in redis).

Still looks good to you?

lib/vps_util.py
```diff
@@ -67,36 +69,46 @@ def highstate_pid(name):
...
-                    time.sleep(10)
+            trysaltcmd('reboot')
+            for _ in xrange(10):
+                time.sleep(10)
+                print("Fetching access data...")
+                try:
+                    adstr = subprocess.check_output(['salt', name, 'cmd.run', 'cat /home/lantern/access_data.json'])
+                except subprocess.CalledProcessError:
+                    continue
```
oxtoacart: Might be worth logging the error text.

aranhoide: Yes, will do.

oxtoacart:
You might want to log the subprocess exception when we can't cat access_data.json, just in case there's something useful in there for debugging.  Otherwise, LGTM and you can pull whenever you like.

aranhoide:
@oxtoacart: thanks for your review!

aranhoide:
Whoops, I merged without pushing the printout of the subprocess exception.  I'll just push that to master.

Issue #140, opened 2016-04-20
Move away users from closed proxies before retiring them for bandwidth consumption 


aranhoide:
When a proxy closes itself because it's hitting its monthly quota, we don't want to put all its users back in the slice table, where they are more vulnerable to infiltration and may cause imbalances.  Since the original server worked alright and was not too underutilized, a good option is to just move all its users to a server freshly pulled from the queue.

This is especially important now that we have some dedicated servers.  But this also points to a balancing issue: pouring the users of a big server into a smaller or bigger one is probably suboptimal at best.

myleshorton:
> But this also points to a balancing issue: pouring the users of a big server into a smaller or bigger one is probably suboptimal at best.

How does #149 address that issue @aranhoide, if at all? How many dedicated servers do we have at this point?

aranhoide:
#149 doesn't address this issue.  We have 4 dedicated servers.  They won't hit their transfer limit this month. My current plan is to offload their users manually to regular proxies. We're not seeing the kind of instability that prompted me to give these a try.

  
If and when we decide to use dedicated proxies for free users too, I don't suppose we'll keep non-dedicated ones around, so I don't see much of a point in addind logic to handle both at the same time.  We can add it when it becomes clear we'll use it.

aranhoide:
@myleshorton correction: we used to have 5, not 4, dedicated proxies, and one of them, launched last April 20, was likely to hit bandwidth limits by the end of the month.

I've offloaded three of them, including this dangerous one.  I'll offload the other two after the server queues recover a bit (it takes some 10 regular proxies to serve the users of a dedicated one).

aranhoide:
All dedicated proxies are offloaded and retired now.

PR #149, opened 2016-04-27
Issue140

aranhoide:
Implements #140.

Tested following this script:

```org
**** Servers start up correctly in test cloudmaster
Just update, highstate, and service status for offload.  Don't care much about the rest at this point.
**** Set up
***** One proxy in the srv tables which I can make open or closed, to test the two paths
use pull_from_srvq without replacement
name='fp-https-doams3ara-20160425-011', ip='178.62.221.149', srv=1
***** One proxy in ir:srvq, to pull in order to take the users of the previous one
***** A backup of such ir:srvq, so we can reset it for further testing
***** An ir:slices table with the first proxy in it, to test that open proxies won't be offloaded
***** An ir:clientip->srv table with some users assigned to the first proxy (and others assigned to bogus server IDs)
***** Kill the retire server, so we can see and reset the entries in the retire queue as we test various scenarios
but note that highstate will restart it!
***** Edit check_vultr_transfer so it will retire the proxy unconditionally
the logic to determine whether a proxy should be retired at all is working well in production and isn't affected by this change
**** Open servers don't offload themselves
***** add an entry to ir:slices with the proxy's srvid as the key
***** manually run the check_vultr_transfer.py script, watching syslog
***** Verify that
****** the relevant printout is displayed
****** the server is enqueued for retirement
****** the ir:clientip->srv table doesn't change at all
****** the replacement proxy is still in ir:srvq
**** Closed servers offload themselves
***** delete the entry for the proxy in retireq
***** remove the entry for the proxy in ir:slices
***** manually run check_vultr_transfer.py again
***** verify that:
****** the relevant printout is displayed
****** the server is enqueued for retirement
****** the server in ir:srvq has been pulled from the queue
****** the users from the original server are assigned to the new one
****** no other users have been touched
```

### DIFF
--- a/lib/vps_util.py
+++ b/lib/vps_util.py
@@ -317,10 +317,10 @@ def retire_proxy(name=None, ip=None, srv=None, reason='failed checkfallbacks', p
         return
     p = pipeline or redis_shell.pipeline()
     if offload:
-        qname = 'offloadq'
+        qname = '%s:offloadq' % region_by_name(name)
     else:
-        qname = 'retireq'
-    p.rpush('%s:%s' % (cm_by_name(name), qname), '%s|%s' % (name, ip))
+        qname = '%s:retireq' % cm_by_name(name)
+    p.rpush(qname, '%s|%s' % (name, ip))
     log2redis({'op': 'retire',
                'name': name,
                'ip': ip,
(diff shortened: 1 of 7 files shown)


aranhoide:
Offloading is actually performed from the cloudmasters in the proxy's region.  The proxy requests being offloaded (which entails being retired) by pushing a message to a reliable queue, just like for retiring.  This is necessary because offloading proxies is memory intensive, and in my preliminary tests a production proxy would swap out like crazy and get from <20% CPU to a pegged one for minutes.

aranhoide:
Sorry about the org formatting in the test script.

aranhoide:
Merging and deploying as explained in email to dev.  Next I'll offload the dedicated servers into regular ones and retire them.

PR #153, opened 2016-05-02
whitelist should never reduce ttl

### LINKED ISSUE #145: "Permanent" IP whitelisting gets reset if you use hussh etc.
Currently, permanent whitelisting will be cleared if you use `hussh` or any command that updates your whitelist.  Maybe we should have a different key pattern for permanent whitelists, so it won't be overwritten by hussh et al, and have sshalert.py check for that as a fallback.

aranhoide:
Fixes #145.  

Tested locally:

- IP permanently whitelist,

- no such IP in whitelist,

- key with lower ttl than the one being set, and

- key with higher ttl than the one being set.

### DIFF
--- a/lib/misc_util.py
+++ b/lib/misc_util.py
@@ -64,7 +64,17 @@ def whitelist_ssh(time=60):
         print >> sys.stderr, "No redis access; I won't be whitelisted in this SSH session"
         return
     ip = subprocess.check_output(['dig', '+short', 'myip.opendns.com', '@resolver1.opendns.com']).strip()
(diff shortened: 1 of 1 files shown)


aranhoide:
Merging since I've tested this and it won't become a priority in a while.  I only did it out of shame for having this issue outstanding, but I probably shouldn't, since I too had more important stuff to do.

oxtoacart:
@aranhoide Sorry I forgot about this.  For what little it's worth, it LGTM.

aranhoide:
@oxtoacart : It's OK, as I said it wasn't a high priority problem to begin with.

PR #156, opened 2016-05-10
implement https://github.com/getlantern/secret_lib/issues/12

aranhoide:


aranhoide:
Merging since it's trivial (only removing one if/else branch and the code it uses) and now it's the best time for deployments and compaction jobs.

PR #162, opened 2016-05-16
various fixes from testing in test cloudmaster

aranhoide:


aranhoide:
Merging since I've tested the refactoring of the `requests` requirement and the launching of servers.  This crept up from a trivial change and is a minor helper that's not worth anyone's time.

aranhoide:
Whoops, I had the wrong PR title.

PR #163, opened 2016-05-16
minor refactoring and utility function

aranhoide:
Tested in the REPL.


aranhoide:
Tested with open and closed servers (no baked-in yet, but those can't get retired anyway).

PR #165, opened 2016-05-17
Updated salt configuration to allow talking to encrypted Redis

oxtoacart:


lib/redis_util.py
```diff
@@ -1,6 +1,7 @@
 from collections import namedtuple
 from datetime import datetime
 import os
+import here
```
oxtoacart: @aranhoide I hope that moving here to the lib folder doesn't break encapsulation or anything :)

aranhoide: @oxtoacart if it doesn't break 'here', it's fine.  But it might, since IIRC 'here' works by reading its own location.

What's the impetus for the change?

aranhoide: Note that it may work for the particular case of secrets_path but still break for others, since the relative path of lantern_aws/secret is the same whether you are in bin/ or in lib/.

aranhoide: Then again, I don't remember if I'm using it for other stuff any more.

aranhoide: @oxtoacart ^

oxtoacart: The impetus was that I need its capability in redis_util.  Since it's minimal code, I can just copy and paste it and give each folder its own version.  Right now, it only uses relative paths, and both bin and lib are at the same level, so everything else is relatively the same too, but going forward it could break so probably better to have two of them.

aranhoide: @oxtoacart: We're not uploading the 'lantern_aws/secret' folder to either the master or minions, so redis_util won't be able to read these files in those machines.  I suggest adding a symlink to these, as is done for other secret files (see, for example, how it's done for the files in lantern_aws/salt/pylib/*.py).

aranhoide: Well, these files are not secret, but the technique works just as well for symlinking into the lantern_aws/secret folder.  rsync will follow these symlinks, while git will store them as is.

aranhoide:
@oxtoacart : this is not ready for review yet, right?  (No rush, just checking that I'm not holding this one.)

oxtoacart:
@aranhoide Correct, I'm working on a full version of this including the salt changes, and I'll integrate your feedback about `here` into there.

oxtoacart:
@aranhoide This is ready for review now, though we don't want to actually pull until we're ready to switch to TLS.

I tested by deploying to a cloudmaster and a config-server and making sure that I could connect to an encrypted redis instance by manually overriding REDIS_URL and trying redis_util.

oxtoacart:
All the python stuff just opens TLS connections to RedisCloud directly, whereas the config-server goes through a local stunnel.  In test cloudmasters, config-server bypasses stunnel since the cloudmaster Redis isn't fronted by its own stunnel.

oxtoacart:
@aranhoide I've pulled out the part that activates stunnel for the config servers, so in theory we could deploy this PR as soon as it passes review.

README.md
```diff
@@ -219,9 +219,9 @@ cloudmaster_address = "188.166.40.244"
 
 - and place it in `~/git/lantern_aws/bin/config_overrides.py`.  You probably want to have it saved somewhere else too, since you'll be deleting and restoring `config_overrides.py` to alternate between the production deployment and one or more non-production ones.
 - `cd ~/git/lantern_aws/bin`
-- `./update.py --as-root`
+- `./update.py --as-root` (you only need to run --as-root until you've successfully run state.highstate once)
 - back in the cloudmaster [2]:
-- `salt-call highstate | tee hslog`
+- `salt-call state.highstate | tee hslog`
```
aranhoide: Good catch, thanks!

salt/config_server/init.sls
```diff
@@ -28,3 +57,4 @@ config-server:
     - watch:
         - file: /home/lantern/config-server.jar
         - file: /etc/init/config-server.conf
+        - cmd: restart-stunnel
```
aranhoide: This will trigger a config server restart unconditionally on every Salt update.  Restarting the config server bumps the CPU usage for the first minute or so.

salt/config_server/init.sls
```diff
@@ -5,11 +5,40 @@ cfgsrv-env:
...
+    - name: /etc/default/stunnel4
+    - pattern: "ENABLED=0"
+    - repl: "ENABLED=1"
+    - append_if_not_found: yes
+    - require:
+      - pkg: stunnel4
+
+restart-stunnel:
+  cmd.run:
+    - name: /etc/init.d/stunnel4 restart
```
aranhoide: The idiomatic way to do this, IIUC, would be to have a `stunnel4` service watching whatever files it depends upon.  Server restart is what happens when a service's watches get triggered.  As is, this command will run unconditionally.

salt/pylib/paths.py
```diff
@@ -0,0 +1,4 @@
+# This version of paths.py is used when running scripts on salt-managed machines
```
aranhoide: Nice!

aranhoide:
@oxtoacart: LGTM, although I'd rather try and limit restarts of the config server unless you think it might get tricky to control that.  Then again, most of the time we update Salt in the config server for any reason we'll probably want to restart the service anyway, so feel free to merge as is if you really meant to restart unconditionally.

salt/config_server/init.sls
```diff
@@ -5,11 +5,40 @@ cfgsrv-env:
...
     - name: /etc/environment
     - text: |
-        AUTH_TOKEN={{ pillar['cfgsrv_token'] }}
-        REDISCLOUD_URL={{ pillar['cfgsrv_redis_url'] }}
+        AUTH_TOKEN='{{ pillar['cfgsrv_token'] }}'
+        REDISCLOUD_URL='{{ pillar['cfgsrv_redis_url'] }}'
         PRODUCTION=true
         PORT=62000
 
+stunnel4:
```
aranhoide: Since there's already this Salt entry called `stunnel4`, a way to introduce a service entry, should you want to, would be to change this to:

```
stunnel4:
  pkg.installed: []
  service.running:
    - watch:
       - file: /etc/stunnel/stunnel.conf
       (etc)
```

The `: []` after `pkg.installed` would be required by YAML.  Being able to omit these (essentially allowing a string in place of a dict key) is a silly Salt shortcut.

salt/stunnel4.sls
```diff
@@ -0,0 +1,18 @@
+stunnel4:
```
oxtoacart: @aranhoide I factored stunnel a little bit better and more reusably, and made it a service.running thing instead of a cmd.run thing as you suggested.

salt/stunnel4.sls
```diff
@@ -0,0 +1,18 @@
...
+  pkg.installed:
+    - refresh: True
+
+  service.running:
+    - enable: yes
+    - watch:
+        - file: /etc/default/stunnel4
+        - file: /etc/stunnel/*
```
aranhoide: Oh, I didn't know you could do that.  Nice!

aranhoide: Actually, have you tested this?  I can't find any documentation for the ability to use globs here.

Also, there's only a single `file` state defined with `/etc/stunnel/`, so I don't think a glob buys us much here.  Just in case there's some misunderstanding here, `watch` this doesn't watch the filesystem for changes or do anything that smart.  It only checks the result of the execution of other salt states (and also implicitly sets a dependency, and hence constrains an order, for their execution).

Finally, I'm not sure you can `watch` states that are not defined in the same file or `include`d.

aranhoide: Actually, as these are now defined, I don't think you could use `include`, since `config-server/init.sls` and `stunnel4.sls` depend on each other.  So Salt better be able to resolve dependencies between all files that go into a minion config.

I'll test this independently, to discard any effect of leftover files.

oxtoacart: Cool.  It appears that stunnel is restarting every time, looking into that now.

oxtoacart: Oh, take that back, it only restarted when I changed stunnel_client.conf, so apparently the wildcard does work.

aranhoide:
Installing `python-hiredis` fails for me.  A
```
python-dev:
  pkg.installed
```
might be the only thing needed to fix this.  You can make python-hiredis depend on that or, on anticipation that this will be needed more generally, you may want to add it to `base_prereqs.sls`.

```
                In file included from src/hiredis.c:1:0:
                src/hiredis.h:4:20: fatal error: Python.h: No such file or directory
                 #include <Python.h>
                                    ^
                compilation terminated.
                error: command 'x86_64-linux-gnu-gcc' failed with exit status 1

```

aranhoide:
@oxtoacart :  after adding `python-dev` everything works fine for me.  Unfortunately, the config server service still restarts unconditionally.  I think this is due to service states always triggering updates when watched.  I'll try a workaround for this, one sec...

oxtoacart:
@aranhoide Thanks!

aranhoide:
@oxtoacart : see my latest commit.  Two things to note:

(i) the mutual dependency between config-server and stunnel4 seems definitely not a problem.  It seems `include` is only required when the salt configuration doesn't otherwise specify the required modules.

(ii) I added a (inelegant and convoluted) trick to make config-server restart if, and only iff, any of its own dependencies, or those of stunnel4, require it.  I've also made it depend on stunnel4, so they should restart in order, but just to be on the safe side I've forced the config server to execute `last` (if in the future anything depends on the config server Salt will do the right thing, `order: last` notwithstanding).

aranhoide:
@oxtoacart I've tested this by (i) checking that reapplying salt state doesn't make the config server restart, (ii) adding a blank line to `stunnel_client.conf` makes both the `stunnel4` and `config-server` services restart, and (iii) a new call to highstate doesn't make any of those restart.

salt/stunnel4.sls
```diff
@@ -1,12 +1,20 @@
+stunnel4-deps:
```
aranhoide: I had a comment about this.  It seems I hadn't saved this file.  One sec...

oxtoacart: Very clever.  Thanks for all your help on this!

PR #166, opened 2016-05-17
changes required to upgrade salt to v2015.8.8.2

aranhoide:
Some of these are actually required to launch new cloudmasters, in any Salt version.


aranhoide:
Merging, since I've tested this in a newly launched cloudmaster, for which the old version doesn't work anyway!

There's nothing much to review about the change to v2015.8.8.2 either.  It's mostly a testing and deployment task.

PR #169, opened 2016-05-20
Self-hosted Redis

oxtoacart:


oxtoacart:
Redis setup with master/slave replication works.

salt/redis_server/init.sls
```diff
@@ -0,0 +1,123 @@
...
+  file.managed:
+    - source: salt://redis_server/stunnel_server.conf
+    - user: root
+    - group: root
+    - mode: 644
+    - makedirs: True
+    - require:
+      - pkg: stunnel4
+
+install-python-software-properties:
```
aranhoide: Unnecessary; already done in https://github.com/getlantern/lantern_aws/blob/master/salt/base_prereqs.sls#L17

salt/redis_server/init.sls
```diff
@@ -0,0 +1,123 @@
...
+
+/etc/redis/redis.conf:
+  file.managed:
+    - source: salt://redis_server/redis.conf
+    - user: root
+    - group: root
+    - mode: 644
+    - makedirs: True
+
+{% if pillar.get('master', False) %}
```
aranhoide: 'master' is quite an overloaded term in Salt (and sysadmin) parlance.  A name that ties this more explicitly to redis should be safer.  'redis-master' or 'is-redis-master'?

salt/redis_server/init.sls
oxtoacart: Fixed

salt/redis_server/init.sls
```diff
@@ -0,0 +1,119 @@
...
+    file.append:
+      - name: /etc/security/limits.conf
+      - text: |
+          redis            soft    nofile            128000
+          redis            hard    nofile            128000
+
+redis-server:
+  pkgrepo.managed:
+    - ppa: chris-lea/redis-server
+    - require:
```
aranhoide: No need for this require either, because the `order` entry in base_prereqs makes sure those are run before any states without an explicit `order`.

aranhoide: BTW, I guess this would have broken if `python-software-properties` was not a `pkg` in `base_prereqs`, because it used to be a `cmd` here.  ;)

salt/redis_server/init.sls
oxtoacart: Changed to is_redis_master.

salt/redis_server/init.sls
oxtoacart: Fixed.  BTW, is a requirement for a pkg based on whether or not that pkg state has run, or whether or not that package is installed?

salt/redis_server/init.sls
```diff
@@ -0,0 +1,119 @@
...
+    - makedirs: True
+{% endfor %}
+
+ufw reload:
+  cmd.run:
+    - require:
+      - pkg: ufw
+    - watch:
+      {% for file in rulefiles %}
+      - file: /lib/ufw/{{ file }}
```
aranhoide: An idea for the future is to put this command in a state that will be used by every firewalled machine and have it just watch `- file: /lib/ufw/*`.  But no need to go the extra mile here and now.

salt/redis_server/redis.conf
```diff
@@ -0,0 +1,697 @@
...
+# 1m => 1000000 bytes
+# 1mb => 1024*1024 bytes
+# 1g => 1000000000 bytes
+# 1gb => 1024*1024*1024 bytes
+#
+# units are case insensitive so 1GB 1Gb 1gB are all the same.
+
+# By default Redis does not run as a daemon. Use 'yes' if you need it.
+# Note that Redis will write a pid file in /var/run/redis.pid when daemonized.
+daemonize yes
```
aranhoide: Why daemonize it?  Is this run by upstart?  If not, is it run by something that knows to restart it if it crashes?

oxtoacart: Good catch!  It's run by SysVinit, so it shouldn't be daemonized.  I'll fix.

aranhoide: Does SysVinit know to restart the redis server if it crashes?

salt/redis_server/init.sls
aranhoide: IIUC, the former.

aranhoide:
@oxtoacart: done with my first round of review (including your fixes up to as of this writing).  All LGTM, barring my doubt about SysVinit being able to restart a crashed service.  I think that would be an important feature until we have failover.

I'll do a deployment from a test cloudmaster when this is final, to check that this PR doesn't inadvertently rely on any files or state local to your dev machine or your cloudmaster (that doesn't seem the case).

oxtoacart:
@aranhoide Thanks for the review!  I'm going to look into whether I can switch to using upstart for redis.

aranhoide:
Some quick googling turned up this

https://www.digitalocean.com/community/tutorials/how-to-configure-a-linux-service-to-start-automatically-after-a-crash-or-reboot-part-1-practical-examples

So it seems you can have SysVinit-managed services automatically restart on crash, but it takes some work.  Maybe the redis-server .deb sets this up?  If not, then yes, upstart might be easiest.

salt/redis_server/redis-server.conf
```diff
@@ -0,0 +1,11 @@
+# redis-server - upstart job file
```
oxtoacart: @aranhoide Now using upstart for redis.

salt/redis_server/init.sls
```diff
@@ -88,12 +88,26 @@ redis-server:
     - refresh: True
     - version: 3:3.0.7-1chl1~trusty1
 
+disable-redis-server-sysv:
```
oxtoacart: @aranhoide This is a little ugly but I think it works to remove the init.d configuration.

salt/redis_server/init.sls
```diff
@@ -88,12 +88,26 @@ redis-server:
...
+  cmd.run:
+    - name: /etc/init.d/redis-server stop ; update-rc.d redis-server disable ; rm /etc/init.d/redis-server
```
aranhoide: Have you checked that `/etc/init.d/redis-server stop` won't kill the redis started by upstart?  It shouldn't, if it relies on pid files...

oxtoacart: Even if it did, upstart will respawn.  Also, this will fail after we've deleted /etc/init.d/redis-server, so it becomes moot after the first time.

README.md
```diff
@@ -229,6 +229,26 @@ Your cloudmaster should be ready now.  If it's not a production one (XXX: add in
...
+`is_redis_master: "True"`.
+
+To launch a redis server named `redis-donyc3-001` in the `donyc3` datacenter:
+
+On the cloudmaster `cm-donyc3`:
+
+```
+sudo touch /srv/pillar/redis-donyc3-001
+# If you want this to be a redis master
+sudo echo 'is_redis_master: "True"' > /srv/pillar/redis-donyc3-001
```
aranhoide: `is_redis_master: true` (or, equivalently, `yes`) would be better, because this may set false expectations about what will `is_redis_master: "False"` do.

aranhoide: Also, have you tested this will work with sudo?  IIRC the redirect will fail in this case, unless you have root permissions pre sudo.

oxtoacart: You mean not quoting "True"?  For some reason, when I used `true` I got an error about it not being iterable or something like that.

salt/config_server/init.sls
```diff
@@ -13,6 +13,10 @@ cfgsrv-env:
 /etc/stunnel/stunnel_client.conf:
   file.managed:
     - source: salt://config_server/stunnel_client.conf
+    - template: jinja
+    - context:
+        redis_host: {{ pillar['cfgsrv_redis_url'].split('@')[1] }}
```
aranhoide: Sourcing this from the redis URL is really neat!

README.md
oxtoacart: Good catch, fixed the sudo issue.

salt/config_server/init.sls
oxtoacart: Thanks!

salt/http_proxy/config.ini
```diff
@@ -13,7 +13,9 @@ logglytoken = {{ pillar['loggly_token'] }}
 pprofaddr = localhost:4000
 proxied-sites-sample-percentage = 0.0001
 proxied-sites-tracking-id = UA-21815217-16
-redis = {{ pillar['secondary_redis_url'] }}
+redis = {{ pillar['cfgsrv_redis_url'] }}
```
aranhoide: LGTM.  This doesn't seem conductive to using the to-be-production DB before deploying it, so we decided we wouldn't do that, right?  (If so, I agree.)

oxtoacart: Oh yeah, good point.  Let me worry about that when I do the changes to enable metrics on http-proxy.  I'll probably do it a little differently, specifically just for the proxies.  I just wanted to get rid of this concept of "secondary redis" from the top-level since it caused so much confusion.

oxtoacart: I'll do that in a separate PR.

README.md
aranhoide: Weird.  I'll try that.

aranhoide:
LGTM.  Testing...

oxtoacart:
@aranhoide Thanks for the thorough review.  Since we're setting this up in our staging environment, that also gives us an opportunity to test.

aranhoide:
@oxtoacart true, but if you deploy the staging setup there's the chance that some hidden dependency on something local to your dev machine sneaks in.  Also, I get some first contact with this stuff. :)

aranhoide:
LGTM.  There's one minor issue with the initial highstate hanging up for redis machines, but merging anyway since work on this area will continue anyway and this doesn't seem like it'll affect non-redis machines.

aranhoide:
I'll keep the branch around for the fix to this highstate issue.

oxtoacart:
@aranhoide Thanks for the really thorough review!

Issue #160, opened 2016-05-12
Add Linode Singapore and Tokyo Datacenters for Pro Users


oxtoacart:

oxtoacart:
One of the core benefits of pro is that users get multiple servers.  If only one of those servers is actually fast, this is not a very good benefit.  Adding Linode Singapore allows us to give Chinese users two fast servers at different datacenters.

PR #170, opened 2016-05-20
Linode vpss (except bandwidth management)

aranhoide:
This was tested in new cloudmasters in Linode Tokyo and Singapore.

- Launching, destroying proxies via the cloudmaster services.

- Running check_vpss (exercises all_vpss).

I think this is a good end-to-end test of all functionality added.

It didn't make sense to try and adapt the Vultr per-proxy code for bandwidth management, since bandwidth is pooled for the whole account in Linode.  This will be done in a separate PR.


aranhoide:
@uaalto assigned to you because I caught you asking @oxtoacart for work to do.  :P

uaalto:
That was about setting up Redis reporting today ;)
Is this urgent? I have a few things in the backlog ATM

aranhoide:
@uaalto I guess it's not any more or less urgent than anything else in the sprint.  It implements this

https://github.com/getlantern/lantern_aws/issues/160

Feel free to reassign if you know anyone less busy.

(cc: @myleshorton, maybe you have a better overview of who seems more likely to have the bandwidth to review this)

aranhoide:
When merging this, `lisgp1` and `litok1` need to be `sadd`ed to `cloudmasters` in redis.

aranhoide:
@oxtoacart maybe you can take a look at this one while I review your redis PR?

lib/do_util.py
```diff
@@ -53,7 +53,7 @@ def destroy_vps(name):
                         headers={"Authorization": "Bearer " + do_token})
     except KeyError:
         print >> sys.stderr, "Droplet not found:", name
-    os.system('salt-key -yd' + name)
+    os.system('salt-key -yd ' + name)
```
aranhoide: Opportunistic unrelated fix.

uaalto:
I'll be happy to do it! Let me make sure proxies are reporting to Redis
(working on it with ox ATM) and I'll get back to this.


Ulysses Aalto
- PGP - https://pgp.mit.edu/pks/lookup?op=get&search=uaalto@getlantern.org

salt/checkfallbacks/init.sls
```diff
@@ -1,7 +1,7 @@
 {% from 'install_from_release.sls' import install_from_release %}
 
 {# define cmd: checkfallbacks-installed #}
-{{ install_from_release('checkfallbacks', '2.1.1') }}
+{{ install_from_release('checkfallbacks', '2.1.1', 'b3d988327f51ba752ab03edb22b30c558ecf7144') }}
```
aranhoide: Opportunistic unrelated optimization.  Waiting for a checkfallbacks download from Singapore or Tokyo for every update was driving me crazy..

salt/install_from_release.sls
```diff
@@ -1,9 +1,10 @@
-{% macro install_from_release(name, version) %}
+{% macro install_from_release(name, version, sha) %}
```
aranhoide: Part of the unrelated optimization referred to above.

salt/pylib/py.bash
```diff
@@ -0,0 +1,3 @@
+#!/bin/bash
+
+python -i -c "import redis_util as ru; import vps_util as vu; import misc_util as mu; r = ru.redis_shell"
```
aranhoide: I have this as an alias locally (where it imports secret_lib as sl too).  I found it surprisingly helpful.

oxtoacart:
@aranhoide I guess we should update inallcms to include the 2 new ones?

oxtoacart:
Code LGTM other than the missing `inallcms` changes.  Once I'm done with my redis stuff, I'll reverse merge into here from master to resolve the merge conflicts and then pull.

aranhoide:
Good catch on inallcms!

Sorry, I hadn't noticed the conflicts.  I can merge this stuff too.  I don't think this is urgent, unless we want to use Linode for Free.

oxtoacart:
I'm touching some of the same code for the Redis stuff, so I figured it's easiest to wait for that to stabilize in staging and then merge.  Sound good?

aranhoide:
@oxtoacart sounds good!

bin/update.py
```diff
@@ -126,13 +127,16 @@ def upload_pillars(as_root=False):
             ' && echo "cloudmaster_name: %s" >> global.sls '
             ' && echo "do_token: %s" > do_credential.sls'
             ' && echo "vultr_apikey: %s" > vultr_credential.sls'
+            ' && echo "linode_password: \'%s\'" > linode_credential.sls '
```
uaalto: Probably not the concern of this PR, but now that you are adding a new VPS provider, perhaps is time to think about a more "templatized" way of generating these echo commands, considering that they are all very similar in structure and seem to be growing.

bin/util.py
```diff
@@ -25,6 +25,11 @@ def read_vultr_credential():
                              ['api-key'])[0]
 
 @memoized
+def read_linode_credential():
```
uaalto: Similar thing applies here, but with a little personal POV that I remember when playing with this code: what about separating all these credentials into a specific module for it. Seems more natural than keeping them in the general VPS module.

lib/linode_util.py
```diff
@@ -0,0 +1,57 @@
...
+li = linode.api.Api(key=os.getenv("LINODE_APIKEY"))
+li_tokyo = linode.api.Api(key=os.getenv("LINODE_TOKYO_APIKEY"))
+
+
+def create_vps(name, req={}, plan=None):
+    vps_util.save_pillar(name, req)
+    if plan is None:
+        plan = vps_util.dc_by_cm(vps_util.my_cm()) + "_1GB"
+    out = subprocess.check_output(["salt-cloud", "-p", plan, name])
+    # Uberhack: XXX update with salt version...
```
uaalto: I don't understand what the Uberhack is doing!

lib/linode_util.py
```diff
@@ -0,0 +1,57 @@
...
+    name = d['name']
+    ip = d['ip']
+    if not vps_util.highstate_pid(name):
+        print("Highstate not running yet; waiting for a bit just in case...")
+        time.sleep(10)
+    while vps_util.highstate_pid(name):
+        print("Highstate still running...")
+        time.sleep(10)
+    print("Highstate done!")
+    return vps_util.hammer_the_damn_thing_until_it_proxies(name)
```
uaalto: Love it! :) Seen it before in the code but didn't have the chance to comment.

lib/linode_util.py
```diff
@@ -0,0 +1,57 @@
...
+    ret = os.system('salt-cloud -yd ' + name)
+
+def all_vpss():
+    ipbyid = defaultdict(list)
+    for l in li.linode_ip_list() + li_tokyo.linode_ip_list():
+        if l['ISPUBLIC'] == 1:
+            ipbyid[l['LINODEID']].append(l['IPADDRESS'])
+    for k, v in ipbyid.items():
+        if len(v) != 1:
+            print >> sys.stderr, "*** WARNING: linode %s has %s public IPs. ***"
```
uaalto: Where are the %s extracted from?

uaalto:
I saw this was reviewed by Ox when I came to send some comments, since I reviewed this offline. Anyway, my comments are more general regarding the structure in general, so less practical for this PR.

bin/update.py
aranhoide: Yes, I've thought about it a few times, but it never became a priority (see the issues list for this project; I guess some of them will seem more worth our time).  This is ugly to the eye but it works.  It seldom causes any problems.  Quoting and escaping bites me sometimes, but I imagine this might become worse with a more templatized thing.

bin/util.py
aranhoide: The breakdown in files here is dumb: there are executables, and any code they share (but is not deployed) goes to `util.py`.  This file is small enough that this works, and it has the advantage that you don't have to think about where to put your new shared function.  What gets deployed (which usually is also code that I often import in the REPL during ops) goes to `lib`, where some more thought is given to module breakdown.

lib/linode_util.py
aranhoide: It parses the output of the `salt-cloud` command to find the IP.  It's a hack because I don't think the output is guaranteed to stay identical across Salt versions, so this is one thing to check (that said, there is similar code in `do_util.py` that kept working untouched in the upgrade from v2015.5.5 to v2015.8.8.2).

The "proper" way to do this would be importing the relevant salt-cloud modules (which are Python) and calling their functions.  In practice, I have observed this to be more problematic and even more fragile than using `salt-cloud` in the command line and parsing its output.  Maybe that's changed in recent Salt versions, but I've stopped checking (scalded cat flees from cold water, as they say).

lib/linode_util.py
aranhoide: Proxies in `salt-cloud`-enabled providers (i.e. DO and Linode) are probably already set up by the time this is called, so this is mostly a homage to Vultr. :)  That function kind of explains why nobody has, to my knowledge, made a salt-cloud adapter for them.

lib/linode_util.py
aranhoide: Heh, from nowhere, good catch.  This is defensive code which I don't know whether will ever be executed, that's why it's untested.  Will fix.

aranhoide:
@uaalto, thanks for your comments!  If any of the stylistic warts in the `bin/` files really bothers you and you want to make them prettier, I'll be happy to review and merge that.

uaalto:
Thanks for the fix @aranhoide! My other suggestions are fairly superficial and cosmetic, and debatable, so we can just leave them for the time being.

oxtoacart:
@aranhoide I resolved a couple of merge conflicts, one in salt/pylib/init.sls and one in update.py.  Do you mind taking a quick look and making sure those still look good to you?

bin/update.py
```diff
@@ -132,11 +131,10 @@ def upload_pillars(as_root=False):
...
-            ' && echo "secondary_redis_url: \"%s\"" >> secondary_redis_credential.sls'
             ' && echo "github_token: %s" > github_token.sls'
             ' && echo "loggly_token: %s" > loggly_token.sls'
-            r' && echo "base: {\"*\": [salt, global], \"fp-*\": [cfgsrv_credential, vultr_credential, secondary_redis_credential, github_token, loggly_token], \"cm-*\": [do_credential, vultr_credential, linode_credential, cfgsrv_credential], \"cs-*\": [cfgsrv_credential]}" > top.sls '
-            ' && sudo mv salt.sls global.sls top.sls do_credential.sls vultr_credential.sls linode_credential.sls cfgsrv_credential.sls secondary_redis_credential.sls github_token.sls loggly_token.sls $(hostname).sls /srv/pillar/ '
+            r' && echo "base: {\"fp-*\": [cfgsrv_credential, vultr_credential, github_token, loggly_token], \"cm-*\": [do_credential, vultr_credential, linode_credential, cfgsrv_credential], \"cs-*\": [cfgsrv_credential], \"*\": [global, salt]}" > top.sls '
+            ' && sudo mv salt.sls global.sls top.sls do_credential.sls vultr_credential.sls cfgsrv_credential.sls github_token.sls loggly_token.sls $(hostname).sls /srv/pillar/ '
```
aranhoide: @oxtoacart : moving `linode_credential.sls` to `/srv/pillar` has been lost to regression here.

aranhoide:
@oxtoacart : done reviewing the merge.  I found a single regression in handling Linode credentials.

bin/update.py
oxtoacart: Fixed, thanks for catching that.

PR #172, opened 2016-05-20
Enabling reporting from proxies to redis.getlantern.org

oxtoacart:
@aranhoide @uaalto This is mostly for bookkeeping, I'm going to pull myself.


salt/http_proxy/config.ini
```diff
@@ -1,23 +0,0 @@
-{% if obfs4_port != 0 %}
```
oxtoacart: @aranhoide @uaalto While we're in a transitional state of having the proxies report to redis.getlantern.org while the rest of the infrastructure uses the old redis, I just moved the config.ini to too-many-secrets and manually put the redis url in there.

See https://github.com/getlantern/too-many-secrets/commit/dca0868a4c57ba30f0d9d84abf7aea0a4758f5ba#diff-ee952044e200d0356ded3d10eca494d9R16

uaalto: Sounds ok, but I don't understand how this help. Just curious...

PR #177, opened 2016-05-24
Preparing for redis migration in staging environment and production

oxtoacart:


.gitignore
```diff
@@ -14,3 +14,4 @@ bin/status
 .#*
 venv
 .venv
+salt/config_server/config-server.jar
```
oxtoacart: This allows me to keep config-server.jar in my local folder for pushing with update.py (at least to non-production environments)

README.md
```diff
@@ -244,11 +244,26 @@ On the cloudmaster `cm-donyc3`:
 ```
 sudo touch /srv/pillar/redis-donyc3-001.sls
 # If you want this to be a redis master
-sudo /bin/sh -exec "echo 'is_redis_master: \"True\"' > /srv/pillar/redis-donyc3-001"
+sudo /bin/sh -exec "echo 'is_redis_master: \"True\"' > /srv/pillar/redis-donyc3-001.sls"
```
oxtoacart: Fixed typo

README.md
```diff
@@ -244,11 +244,26 @@ On the cloudmaster `cm-donyc3`:
...
+sudo rm /var/lib/dpkg/lock
+sudo dpkg --configure -a
+sudo apt-get install redis-server=3:3.0.7-1chl1~trusty1
+sudo rm /var/cache/salt/minion/proc/*
+```
+
+After this point, run `state.highstate` again and it should run successfully
+from here on out. The machine may still be running a Redis instance started by
+the old init.d script. You can fix this by rebooting the machine.
+
```
oxtoacart: This is annoying, but it works and we don't launch redis servers that often.

bin/update.py
```diff
@@ -101,26 +101,35 @@ def upload_pillars(as_root=False):
...
+    redis_domain = redis_host.split(":")[0]
+    redis_via_stunnel_url = cfgsrv_redis_url.split('@')[0].replace("rediss", "redis") + "@localhost:6380"
+
+    if util.in_dev():
+        redis_host = "%s:6379" % config.cloudmaster_address
+        cfgsrv_redis_url = "redis://redis:%s@%s" % (cfgsrv_redis_test_pass, redis_host)
+        redis_domain = "redis-staging.getlantern.org"
+        # Bypass stunnel in dev environments because we're not encrypting connections to Redis
+        redis_via_stunnel_url = cfgsrv_redis_url
+
```
oxtoacart: New logic for determining redis URLs depending on the environment.

bin/update.py
```diff
@@ -101,26 +101,35 @@ def upload_pillars(as_root=False):
...
             r' && echo "include: [{{ grains[\"id\"] }}]" >> salt.sls '
             ' && echo "" > $(hostname).sls""'
+            ' && echo "in_dev: %s" >> global.sls '
             ' && echo "in_staging: %s" > global.sls '
             ' && echo "in_production: %s" >> global.sls '
             ' && echo "datacenter: %s" >> global.sls '
             ' && echo "cfgsrv_redis_url: %s" >> global.sls'
+            ' && echo "redis_via_stunnel_url: %s" >> global.sls'
+            ' && echo "redis_host: %s" >> global.sls'
+            ' && echo "redis_domain: %s" >> global.sls'
```
oxtoacart: Putting this stuff in global.sls means I'm doing less logic in the Jinja templates.

salt/redis.sls
```diff
@@ -3,6 +3,16 @@ redis-env:
...
+# In case we're switching back and forth between redis configurations, replace
+# what may have already been in /etc/environment.
+replace-redis-env:
+  file.replace:
+    - name: /etc/environment
+    - pattern: "REDIS_URL=.+"
+    - repl: "REDIS_URL={{ pillar['cfgsrv_redis_url'] }}"
+    - require:
+      - file: redis-env
+
```
oxtoacart: This took me a while to figure out.  I had prematurely switched the Redis URL at one point.  When I wanted to finally switch to that URL, I couldn't because it was no longer getting appended (because append only appends if it doesn't see the text in the file already).  Solution was to replace existing REDIS_URL statements so that we have multiple that all say the same thing.

salt/redis_server/init.sls
```diff
@@ -14,7 +14,11 @@
 
 /etc/redis/redis_auth.conf:
   file.managed:
+    {% if pillar["in_production"] %}
     - source: salt://redis_server/redis_auth.conf
+    {% else %}
+    - source: salt://redis_server/redis_auth_test.conf
+    {% endif %}
```
oxtoacart: In staging we use a special hardcoded password of 'testing'.

salt/redis_server/init.sls
```diff
@@ -98,7 +102,7 @@ redis-server:
 
 disable-redis-server-sysv:
   cmd.run:
-    - name: /etc/init.d/redis-server stop ; update-rc.d redis-server disable ; update-rc.d redis-server remove ; rm /etc/init.d/redis-server ; echo "done"
+    - name: /etc/init.d/redis-server stop && update-rc.d redis-server disable
```
oxtoacart: This was an attempt to leave Redis in a somewhat better state after first start.  Not sure it did much, but it didn't hurt.

salt/redis_tls/init.sls
```diff
@@ -1,4 +1,6 @@
-{% set files=['client_key.pem', 'client_cert.pem', 'redis.getlantern.org/fullchain.pem', 'redis.getlantern.org/cert.pem', 'redis.getlantern.org/privkey.pem'] %}
+{% set files=['client_key.pem', 'client_cert.pem',
+              'redis.getlantern.org/fullchain.pem', 'redis.getlantern.org/cert.pem', 'redis.getlantern.org/privkey.pem',
+              'redis-staging.getlantern.org/fullchain.pem', 'redis-staging.getlantern.org/cert.pem', 'redis-staging.getlantern.org/privkey.pem'] %}
```
oxtoacart: Added certs for redis-staging.getlantern.org.

salt/sshalert/sshalert.py
```diff
@@ -17,7 +17,11 @@
     ip = os.environ['SSH_CONNECTION'].split(' ')[0]
     user = os.environ['USER']
 
-    whitelisted = redis_shell is not None and redis_shell.exists('sshalert-whitelist:%s' % ip)
+    try:
+        whitelisted = redis_shell is not None and redis_shell.exists('sshalert-whitelist:%s' % ip)
+    except Exception,e:
+        print >> sys.stderr, "Unable to check whitelisted status, assuming whitelisted: ", e
+        whitelisted = True
```
oxtoacart: During the migration, I "disable" redis by changing the password.  Unfortunately, that cases redis_shell.exists() to throw an exception.  Rather than barfing on this, I just assume that whitelisted is True since this is just temporary.

salt/top.sls
```diff
@@ -42,7 +42,7 @@ base:
     'cs-*':
         - lantern_build_prereqs
         - proxy_ufw_rules
-#        - stunnel4
+        - stunnel4
```
oxtoacart: Reenabling stunnel for config servers.

README.md
```diff
@@ -244,11 +244,26 @@ On the cloudmaster `cm-donyc3`:
...
 sudo salt "redis-donyc3-001" state.highstate
 ```
 
+The first time you run highstate, redis may fail to install. To resolve this,
+ssh to the machine as lantern and then run the following:
+
+```
+ssh lantern@<ip>
+sudo rm /var/lib/dpkg/lock
+sudo dpkg --configure -a
```
aranhoide: This step failed for me last time I tried.  I'll assume this works as of this PR.

bin/update.py
```diff
@@ -101,26 +101,35 @@ def upload_pillars(as_root=False):
...
-            redis_address = config.cloudmaster_address
-        cfgsrv_redis_url = "redis://redis:%s@%s:6379" % (cfgsrv_redis_test_pass,
-                                                         redis_address)
+
+    if util.in_staging():
+        cfgsrv_redis_url = "rediss://:testing@redis-staging.getlantern.org:6380"
+
+    redis_host = cfgsrv_redis_url.split('@')[1]
+    redis_domain = redis_host.split(":")[0]
+    redis_via_stunnel_url = cfgsrv_redis_url.split('@')[0].replace("rediss", "redis") + "@localhost:6380"
```
aranhoide: I'd rather have an explicit test for `util.in_production()` than risking pushing production credentials to test cloudmasters if other checks give wrong results in the future.  Since these three are mutually exclusive, maybe we can have an `elif` chain that raises an exception if none of the three are met:

```python
if util.in_production():
   ...
elif util.in_staging():
   ...
elif util.in_dev():
   ...
else:
   assert False, "Exactly one of the above must be true!"
```

The original code was more or less at fault in this regard too, so I can make this change if you're too busy.

bin/update.py
aranhoide: LGTM.

bin/util.py
```diff
@@ -14,6 +14,9 @@ def in_production():
 def in_staging():
     return config.cloudmaster_name.endswith('staging')
 
+def in_dev():
+    return not in_production() and not in_staging()
```
aranhoide: I like making this more explicit, thanks.

README.md
oxtoacart: Yeah, I followed these steps when setting up the 2nd staging redis.

salt/redis.sls
aranhoide: It seems cleaner and more expressive to delete the obsolete lines:
```
delete-obsolete-redis-env:
   file.line:
      - name: /etc/environment
      - match: "REDIS_URL=(?!{{ pillar['cfgsrv_redis_url' }}).*"
      - mode: Delete
```
This is untested.  I'm going by the `state.line` docs from https://docs.saltstack.com/en/latest/ref/states/all/salt.states.file.html.

bin/update.py
oxtoacart: Fixed.

salt/redis_server/init.sls
```diff
@@ -3,8 +3,8 @@
     - source: salt://redis_server/stunnel_server.conf
     - template: jinja
     - context:
-        redis_host: {{ pillar['cfgsrv_redis_url'].split('@')[1] }}
-        redis_domain: {{ pillar['cfgsrv_redis_url'].split('@')[1].split(":")[0] }}
+        redis_host: {{ pillar['redis_host'] }}
+        redis_domain: {{ pillar['redis_domain'] }}
```
aranhoide: Yeah, much saner. :)

salt/redis_server/init.sls
aranhoide: This will apply to both staging and testing, right?  Is that OK?

salt/redis_tls/init.sls
aranhoide: Is every file pushed to every VPS?  It would seem safest not to push production certs to non-production machines, because security issues are more likely to accidentally hit VPSs while we're testing new stuff.

salt/sshalert/sshalert.py
aranhoide: `traceback.print_exc(file=sys.stderr)` is advisable if you're using a catch-all exception handler.

salt/redis.sls
oxtoacart: Good idea, I've applied it (with some modifications after testing).

salt/redis_server/init.sls
oxtoacart: The assumption is that we don't launch independent redis servers in testing, so it's moot.

.gitignore
aranhoide: You can just push once to `/home/lantern` for cloudmasters where you want to launch config servers (see `config-server/script/deploy.sh`).

How are you preventing your local copy of config-server.jar from overriding the one in production environments, or how are you preventing this file from being uploaded to all cloudmasters when using `bin/inallcms`?

salt/redis.sls
```diff
@@ -1,17 +1,18 @@
...
-# In case we're switching back and forth between redis configurations, replace
-# what may have already been in /etc/environment.
-replace-redis-env:
-  file.replace:
-    - name: /etc/environment
-    - pattern: "REDIS_URL=.+"
-    - repl: "REDIS_URL={{ pillar['cfgsrv_redis_url'] }}"
     - require:
-      - file: redis-env
+      - file: clean-redis-env
```
aranhoide: I don't think it matters in which order you run these.  Can't hurt, of course.

salt/redis_server/init.sls
aranhoide: Oh, sure.  Sorry for the brain fart.  :)

aranhoide:
Done with preliminary review.  Please let me know when this is final.

salt/redis_tls/init.sls
oxtoacart: Fixed.

salt/sshalert/sshalert.py
oxtoacart: Fixed.

.gitignore
oxtoacart: Oh good point, I've deleted it locally and removed from gitignore.

salt/sshalert/sshalert.py
```diff
@@ -17,7 +17,11 @@
     ip = os.environ['SSH_CONNECTION'].split(' ')[0]
     user = os.environ['USER']
 
-    whitelisted = redis_shell is not None and redis_shell.exists('sshalert-whitelist:%s' % ip)
+    try:
+        whitelisted = redis_shell is not None and redis_shell.exists('sshalert-whitelist:%s' % ip)
+    except Exception, e:
+        print >> sys.stderr, "Unable to check whitelisted status, assuming whitelisted: ", e
```
oxtoacart: @aranhoide I tried the traceback syntax that you suggested, but that actually caused the script to abort.  I spent a little time debugging, but didn't get anywhere with it, so I went back to this which works.

salt/redis_server/init.sls
```diff
@@ -120,6 +124,45 @@ redis-server-running:
         - file: /etc/init/redis-server.conf
         - cmd: stunnel4-deps
 
+{% if not pillar["in_dev"] and pillar.get("is_redis_master", False) %}
```
oxtoacart: @aranhoide I added s3 backups capability to the master.  Staging and production go into their own folders, secured by their own credentials.

salt/sshalert/sshalert.py
aranhoide: Sorry @oxtoacart : traceback is a regular Python module so it needs to be imported.  Just add `import traceback` along with the others and it should work.

salt/redis_server/init.sls
aranhoide: @oxtoacart nice!

salt/redis.sls
```diff
@@ -1,18 +1,10 @@
...
-
 redis-env:
-  file.append:
+  file.replace:
     - name: /etc/environment
-    - text: "REDIS_URL={{ pillar['cfgsrv_redis_url'] }}"
-    - require:
-      - file: clean-redis-env
+    - pattern: "^REDIS_URL=.+$"
+    - repl: "REDIS_URL={{ pillar['cfgsrv_redis_url'] }}"
```
aranhoide: Missing `append_if_not_found`?

oxtoacart: Yes!

PR #178, opened 2016-05-24
Updated dimensions for borda, new binary

oxtoacart:
This also includes binaries for the config-server.jar (pulled from production donyc3 cloudmaster) and http-proxy (pulled from v.0.0.20 release).


.gitattributes
```diff
@@ -0,0 +1 @@
+salt/borda/borda filter=lfs diff=lfs merge=lfs -text
```
oxtoacart: Added by git lfs.

salt/http_proxy/init.sls
```diff
@@ -34,6 +31,7 @@ fp-dirs:
 
 # To copy verbatim.
 {% set nontemplate_files=[
+    ('/home/lantern/', 'http-proxy', 'http-proxy', 'lantern', 755),
```
aranhoide: What will this do if `http-proxy` is currently running?  The old version wasn't working all the time, but we had got restart error rate very low after some tweaks to the `install_http_proxy.py` script.

aranhoide: Just to be clear, it was not failing 100% of the time before that.  So maybe we just try and see whether the error rate gets better or worse.  This calls for being extra careful to `check_deployment.py` after every http-proxy update.

salt/borda/borda.conf
```diff
@@ -31,4 +31,4 @@ limit nofile 1024768 1024768
 #        moment for security reasons.
 # Note - the file authkey has to be manually populated with the authentication
 #        key.
-exec /home/lantern/borda -config /home/lantern/config.ini -stderrthreshold WARNING 2>&1 | logger -t borda
+exec /home/lantern/borda -config /home/lantern/config.ini -stderrthreshold INFO 2>&1 | logger -t borda
```
aranhoide: How chatty is borda?

aranhoide:
@oxtoacart LGTM, but where is the "updated dimensions" part?  I imagine the change is embodied in the borda binary, right?

aranhoide:
@oxtoacart : I'll try installing lfs and checking that I can deploy this to a test cloudmaster and launch a proxy, then merge.

salt/http_proxy/init.sls
oxtoacart: Interesting.  I just tested, and the server did restart even though none of the config files/binary changed.  Are you saying the old mechanism didn't restart http-proxy unnecessarily?

salt/borda/borda.conf
oxtoacart: I don't think it's particularly chatty.

oxtoacart:
Thanks for the review @aranhoide.  The "updated dimensions" part is in too-many-secrets, and already got merged elsewhere.

salt/http_proxy/init.sls
aranhoide: What I mean is that in the old mechanism (esp. before the abovementioned tweaks), the server would sometimes *fail* to restart, i.e., it would stay dead.  I'm not sure whether restarts to http-proxy would happen all the time or only if strictly needed, but I guess that's less of an issue.

oxtoacart: Oh, interesting.  Well, so far it seems to be restarting okay.

aranhoide:
Tested this git-lfs thing.  Merging...

PR #179, opened 2016-05-24
fix parallel SSH

aranhoide:
For some reason, timeout didn't work in subprocess.check_output when SSHing
into a site that demands a password.

This is needed by secret_lib.interactive_offload.

Tested by using it there.  I also tested that pushing this to my test
cloudmaster worked fine and that subprocess32 could be imported there
afterwards.

### DIFF
--- a/lib/misc_util.py
+++ b/lib/misc_util.py
@@ -59,7 +59,7 @@ def ssh(ip, cmd, timeout=30, whitelist=True):
     if whitelist:
         whitelist_ssh()
     return cmd_error_and_output(
-        ['ssh', '-o', 'StrictHostKeyChecking=no', 'lantern@' + ip, cmd],
+        ['ssh', '-o', 'StrictHostKeyChecking=no', '-o', 'BatchMode=yes', 'lantern@' + ip, cmd],
         timeout)
 
 def whitelist_ssh(time=60):
--- a/lib/misc_util.py
+++ b/lib/misc_util.py
@@ -5,10 +5,16 @@ from functools import wraps
 import re
 import multiprocessing
 import sys
-import subprocess
 import time
 import traceback
-
+try:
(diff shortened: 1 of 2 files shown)


lib/misc_util.py
```diff
@@ -76,17 +82,13 @@ def whitelist_ssh(time=60):
     if ttl < time:
         redis_shell.setex(key, 'admin', time)
 
-def with_timeout(args, timeout=None):
-    if timeout is None:
-        return args
-    else:
```
aranhoide: This wouldn't work, in the context of `subprocess.check_output`, for a SSH session hung up because the remote site demands a password.  `timeout 10 ssh thisuserdoesntexist@some-site-with-ssh-passwords-enabled` in the shell aborts the command correctly after 10 seconds, though.

aranhoide:
@fffw while this fix is important, it's not as urgent to review since I only need this for interactive work, and I can do that from this branch.  Of course, it's best to prevent branch rot.

lib/misc_util.py
fffw: [`ssh -oBatchMode=yes`](http://serverfault.com/questions/61915/how-do-i-make-ssh-fail-rather-than-prompt-for-a-password-if-the-public-key-authe) can complement by failing immediately instead of waiting for timeout.

fffw:
@aranhoide All good to me. The batch mode option could be a plus I think.

lib/misc_util.py
aranhoide: Nice!  This might have done it on its own for this particular case.  I'll add it, and I'll keep the backport of Python 3's `timeout` option to `subprocess.check_output`.

aranhoide:
@fffw definitely a plus!  I'll add it, test, and merge.  Thanks!

PR #180, opened 2016-05-24
Plot device bandwidth

uaalto:
Want to take a look at this @aranhoide ?
This python code might have some nice improvements that you can easily spot.

### DIFF
--- a/etc/device_tx_analysis.py
+++ b/etc/device_tx_analysis.py
@@ -65,12 +65,16 @@ if limit_devices != 0:
 
 mean=total_tx / used_devices
 median=with_rank[len(with_rank)/2][1]
+percentile90=with_rank[int(0.1*used_devices)][1]
+percentile95=with_rank[int(0.05*used_devices)][1]
 
 print "Statistics limited to:", used_devices, "devices with minimum", humanize.naturalsize(min_transfer)
 if min_transfer != 0:
     print discarded_devices, "devices transferred less or equal to", humanize.naturalsize(min_transfer)
 print "Mean:", humanize.naturalsize(mean)
 print "Median:", humanize.naturalsize(median)
+print "95 percentile:", humanize.naturalsize(percentile95)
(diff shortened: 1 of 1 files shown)


etc/device_tx_analysis.py
```diff
@@ -0,0 +1,127 @@
...
+        writer.writerows(data)
+
+
+def clean_data():
+    with open("dump.csv", "rb") as f:
+        reader = csv.reader(f)
+        for device, tx in reader:
+            yield float(tx)
+
+rev = reversed([row for row in clean_data()])
```
aranhoide: `reversed(clean_data())` should work too.

etc/device_tx_analysis.py
```diff
@@ -0,0 +1,127 @@
...
+
+with_rank = []
+total_tx, used_devices, discarded_devices = 0, 0, 0
+for tx in rev:
+    if tx < min_transfer:
+        discarded_devices += 1
+        continue
+    used_devices += 1
+    total_tx += tx
+    with_rank += [[used_devices, tx]]
```
aranhoide: `with_rank.append([used_devices, tx])` would have been more idiomatic.  I'm not positive about this, but it might be faster in general -- not that it matters here.

aranhoide:
So there are three things I could review here:

- Safety to merge into master: sure, this is not used by any automated system.

- Style: added a couple comments, but in general this seems idiomatic, even sophisticated, Python.

- Does it do what it intends?: I'm not familiar with the libraries involved and I don't have the time for an in-depth review and testing.  Merging it anyway since this is not used for mission critical stuff AFAIK.

etc/device_tx_analysis.py
uaalto: Oh, yes that's right. This is because I had some other fields before and this is a leftover

etc/device_tx_analysis.py
uaalto: Thanks!

PR #181, opened 2016-05-25
Bumped up file descriptor limit on config-server

oxtoacart:
I tested this by manually applying this config to a single config-server in production. The server came up correctly and /proc/<pid>/limits showed the right file descriptor limit. Curling my config from that server succeeded as well.


oxtoacart:
LGTM.

Issue #176, opened 2016-05-23
checkfallbacks failing for IPs no longer ours


aranhoide:
This happened last Friday, starting around 7:20PM UTC.  It started with just a couple proxies, and eventually affected some 30 ones.  More details in a dev email thread with subject "30 proxies broken for most of yesterday".

aranhoide:
This used to only happen in Vultr Tokyo, but today I've got some cases in DO NYC3.  They had been launched today.

aranhoide:
I've detected a few entries pointing to no longer existing proxies in the server queues for China and Iran.  That could easily cause this.  https://github.com/getlantern/lantern_aws/issues/185 could too.  All three conditions are now cleaned up from our production DB and automatically checked for every 1/2h.  Whichever kicks in first is unlikely to be caused by the others, and is a likely cause of others.

aranhoide:
Today one sequence happened that could lead to this:

- check_vpss detects some proxies as unused (because they were still being launched and weren't registered),

- I go on to destroy those proxies, after checking:

  * that there were no entries in the server request queue for the cloudmasters and that there were no recent refill_srvq.  I checked other stuff first, so quite some time elapsed between check_vpss and my check of the request queue.  3 proxies got registered in the meanwhile.

  * that there had been no recent proxy registrations.  **Mistake**: see this excerpt from the vltok1 syslog:

```
[...]

PR #189, opened 2016-05-26
Srv2cfg sanity checks

aranhoide:
This is **not** a real fix for https://github.com/getlantern/lantern_aws/issues/176, but it should reduce the harm it causes to users while we fix it, and it will serve as a continuous regression check afterwards. 

- Check that all entries in `srv->cfg` correspond to existing proxies.  Remove all `srv->cfg` (and related) entries where this is not true.  

- Check that the configs in redis match the ones in their local access_data.

- Alert to slack for failures in any of the above.

- Change sanity checks in general to be run every 1/2h.  I used to think this was too expensive, but I've fund that our production redis DB deals fine with this.

### DIFF
--- a/lib/alert.py
+++ b/lib/alert.py
@@ -16,6 +16,12 @@ instance_id = "{{ grains['id'] }}"
 ip = "{{ external_ip(grains) }}"
 url = os.environ.get('SLACK_WEBHOOK_URL')
 
+# Allow for testing in developer machines. You still need to set
+# SLACK_WEBHOOK_URL locally, of course.
+if instance_id.startswith('{{'):
+    import getpass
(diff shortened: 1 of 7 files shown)


aranhoide:
Don't merge yet.  I've tested this locally, but I've just realized it won't work in a cloudmaster since its minions aren't accepting the cloudmaster's key.

aranhoide:
It turns out that the part that needed SSH was not essential to the checks for what we're actually seeing fail in production.  I've disabled these checks and tested again.  I'll test in a test cloudmaster, then merge.

aranhoide:
Works OK in my test cloudmaster.  Merging and pushing to DONYC3.

PR #190, opened 2016-05-26
Made the Proxy usage reports script more decent

uaalto:


etc/device_tx_analysis.py
```diff
@@ -17,111 +18,129 @@
...
+        retrieve_redis = misc_util.confirm("Do you want to update the Redis file?")
+
+    if retrieve_redis:
+        print "Retrieving data from Redis..."
+        data = redis_shell.zrange('client->bytesInOut', 0, -1, withscores=True)
+        with open("dump.csv", "wb") as f:
+            writer = csv.writer(f)
+            writer.writerows(data)
+
+    rev = reversed([r for r in clean_data()])
```
uaalto: This actually didn't work without the list comprehension. Might be because clean_data is a generator (is it?)

uaalto:
Not urgent @aranhoide, but is fairly straightforward

etc/device_tx_analysis.py
aranhoide: Sorry, reversed takes a sequence.  So you can say `reversed(list(clean_data()))`.

aranhoide:
LGTM (I guess you only wanted to ask the reverse thing?).

PR #200, opened 2016-05-29
Checking git lfs status before pushing to production

oxtoacart:


bin/update.py
```diff
@@ -28,6 +34,10 @@ def check_master_if_in_production():
         pull_output = subprocess.check_output(['git', 'pull'])
         if pull_output != EXPECTED_PRODUCTION_GIT_PULL_OUTPUT:
             not_up_to_date()
+        lfs_status_output = normalize_status_output(
+                subprocess.check_output(['git', 'lfs', 'status']))
+        if lfs_status_output != EXPECTED_PRODUCTION_GIT_LFS_STATUS_OUTPUT:
```
aranhoide: Some specific text about setting up `git lfs` might help, but this does the real important job, which is prevent us from deploying from machines without git lfs set up.

PR #205, opened 2016-05-31
overload check

aranhoide:
Test a new way to detect whether a machine is overloaded.  cpu_io is observed primarily, and I'm also reporting swap_tx, should that help confirm our suspicions that a proxy is overloaded.

This will do nothing but report to #staging-alerts ATM, but if it proves a good predictor I'll incorporate this check in the upcoming auto-offload/split PR.

### DIFF
--- a/salt/stats/init.sls
+++ b/salt/stats/init.sls
@@ -9,3 +9,9 @@
   cron.present:
     - identifier: save_stats
     - user: lantern
+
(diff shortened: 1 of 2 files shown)


aranhoide:
Merging since I've tested this in my test cloudmaster, it doesn't really do anything, and it runs as its own cron job where it pretty much can't break anything else.

PR #207, opened 2016-06-01
Added monitoring of redis

oxtoacart:


salt/cloudmaster/check_redis.py
```diff
@@ -0,0 +1,29 @@
...
+    redis_url = os.getenv('REDIS_URL')
+    redis_host = redis_url.split("@")[1]
+    alert.send_to_slack("Redis Unavailable",
+                        msg,
+                        color="#ff0000",
+                        channel="#redis-alerts")
+
+try:
+    from redis_util import redis_shell
+    if redis_shell.ping():
```
aranhoide: Will a read-only slave respond to ping?  I understand that the question may be moot at this point, but for future robustness we may want to set some key to a random value and read to check that we got the same?  It's not much more trouble.

salt/cloudmaster/check_redis.py
```diff
@@ -0,0 +1,29 @@
...
+import sys
+import os
+import alert
+
+redis_url = os.getenv('REDIS_URL')
+redis_host = redis_url.split("@")[1]
+
+def fail(msg):
+    redis_url = os.getenv('REDIS_URL')
+    redis_host = redis_url.split("@")[1]
```
aranhoide: You have the same definitions at the module level.  Why repeat them here?

aranhoide: Also, you're not using them inside this function.

salt/cloudmaster/check_redis.py
```diff
@@ -0,0 +1,29 @@
...
+    alert.send_to_slack("Redis Unavailable",
+                        msg,
+                        color="#ff0000",
```
aranhoide: `color='danger',` does the same and is a bit more expressive.

salt/cloudmaster/check_redis.py
```diff
@@ -0,0 +1,29 @@
+#!/usr/bin/env python
+
+# This script checks whether Redis is available by pinging it and alerts on
+# Slack if it isn't.
+import sys
+import os
+import alert
+
+redis_url = os.getenv('REDIS_URL')
```
aranhoide: `redis_url` is available from `redis_util`, so this is a bit of duplication.  Should we want to change the name of the env var (e.g., if we wanted to introduce other redis URLs for any reason) we'd now need to change this in two places.  Not the end of the world, but unnecessary.

salt/cloudmaster/check_redis.py
```diff
@@ -0,0 +1,29 @@
...
+
+try:
+    from redis_util import redis_shell
+    if redis_shell.ping():
+        print "Redis is Up!"
+    else:
+        msg = "Redis at %s did not respond to ping" % redis_host
+        fail(msg)
+except Exception,e:
+    msg = "Unable to check whether redis at %s is available: %s" % (redis_host, e)
```
aranhoide: Might as well just print the whole URL instead of the host only?  You never know when the extra info could be helpful, and it's less work to get it all.

salt/cloudmaster/check_redis.py
aranhoide: I have sometimes found it useful to print these to stderr too (e.g. `print >> sys.stderr, msg`).

aranhoide:
LGTM.  The only substantial comment I have is that maybe we want to check for writability too?

salt/cloudmaster/check_redis.py
oxtoacart: @aranhoide The whole url contains the password, which I didn't think was appropriate to post to Slack.

salt/cloudmaster/check_redis.py
oxtoacart: Oh, good catch, will fix that.

salt/cloudmaster/check_redis.py
oxtoacart: It's intentionally duplicated in case we fail to import redis_util for some reason.

salt/cloudmaster/check_redis.py
oxtoacart: Thanks, didn't know that.

salt/cloudmaster/check_redis.py
oxtoacart: I don't understand the question.  Whatever server is at our Redis URL will respond to the ping.

oxtoacart:
@aranhoide What would checking writability test that ping doesn't?

salt/cloudmaster/check_redis.py
aranhoide: Good point!

aranhoide:
@oxtoacart that this redis server is working as a master and not a read-only slave.

aranhoide:
@oxtoacart I'm assuming that a read-only slave would pass the ping test, which I'm not sure of actually.

oxtoacart:
@aranhoide Ah, I see your point about testing a write.  That's a good idea, I'll do it.

aranhoide:
@oxtoacart please leave the ping test too; if the write test fails but the ping one succeeds, that seems useful to know.

aranhoide:
@oxtoacart another idea: write to a key that is somehow specific to the tester, so multiple checkers don't clobber each other.

PR #208, opened 2016-06-02
Refactor stats.py and make it a library

aranhoide:
Refactor stats.py so it can be useful to other system utilities (in pacticular, check_load.py is the use case that prompted this), and actually make it a library.

The impetus for this change is to make it convenient and clean to use historical monitoring data to determine whether a proxy is overloaded, and thus should be offloaded and/or split.  I thought it would be easiest and less risky to review and deploy this separately.

`check_load` is a tentative experiment in this direction.  Since it's ineffectual at the moment, it doesn't need review.

I've tested this locally (after scp'ing test data from a production proxy) and tested that I got similar values with the refactored `get_bps` as with the old one.  The small difference may be explained by the fact that the old get_bps discards the first update after a reboot, while the new one correctly adds it.

I also tested the Salt changes in my test cloudmaster.


lib/stats.py
```diff
@@ -0,0 +1,149 @@
...
+import os
+import sys
+
+from dateutil.parser import parse as parse_time
+import psutil
+
+from misc_util import obj
+
+
+num_samples = 60 * 24 * 7  # keep last week, assuming 1m precision
```
aranhoide: Works out to about 1MB in practice, so we could keep more, but I've never needed more.

lib/stats.py
```diff
@@ -0,0 +1,149 @@
...
+
+path = "/home/lantern/stats"
+
+delta_val = namedtuple('accum_val', ['val', 'prev'])
+avg_val = namedtuple('avg_val', ['val', 'nsamples'])
```
aranhoide: There are two major classes of stats handled here: cumulative ones that are defined as a counter since last reboot, and others.  For the former we accumulate by adding deltas (with some care to handle reboots, which reset the stat).  For the latter we average over the requested period.

Actually, I'll add this as a code comment.

lib/stats.py
```diff
@@ -0,0 +1,157 @@
...
+    cpu = psutil.cpu_times_percent(interval=1)
+    duse = psutil.disk_usage('/')
+    dtx = psutil.disk_io_counters(perdisk=False)
+    if os.path.exists(path):
+        lines = file(path).readlines()[-num_samples:]
+    else:
+        lines = []
+    # DRY: stat_defs
+    new_line = "|".join(map(str,
+        [now, la1m, cpu.user, cpu.system, cpu.iowait, cpu.idle, mem.percent, mem.active, swap.percent, swap.sin + swap.sout,
```
aranhoide: I no longer think it's a good idea to lump in + out of some stats, but the motivation to fix this is currently not strong enough to port or discard all old data in proxies.

lib/stats.py
```diff
@@ -0,0 +1,157 @@
...
+
+def summary(stats, minutes_back=None):
+    if minutes_back is None:
+        start_time = datetime.fromtimestamp(0)
+    else:
+        # allow string arguments for command line usage.
+        minutes_back = int(minutes_back)
+        start_time = datetime.utcnow() - timedelta(minutes=minutes_back)
+    defs = map(name2def.get, stats)
+    accum_vals, accum_fns = zip(*(d.accum_type for d in defs))
```
aranhoide: `zip(*seq)` transposes `seq`.  In this case it extracts two lists from a list of pairs.

lib/stats.py
```diff
@@ -0,0 +1,157 @@
...
+
+# XXX: quick check; move to check_load when stats.py becomes a library.
+def check_load(minutes_back=30):
+    s = summary(['cpu_io', 'mem_pc', 'swap_tx'], minutes_back=minutes_back)
+    mem_pc = s['values']['mem_pc']
+    cpu_io = s['values']['cpu_io']
+    swap_tx = s['values']['swap_tx']
+    details = "cpu_io: %.2f, mem_pc: %.2f, swap_tx: %s, num_cores: %s" % (cpu_io, mem_pc, swap_tx, psutil.NUM_CPUS)
+    print details
+    if cpu_io > 3 and mem_pc > 85:
```
aranhoide: Still playing with these numbers; this function does nothing in practice currently, so it doesn't need review.

lib/stats.py
```diff
@@ -0,0 +1,157 @@
...
+
+from dateutil.parser import parse as parse_time
+import psutil
+
+from misc_util import obj
+
+
+num_samples = 60 * 24 * 7  # keep last week, assuming 1m precision
+
+path = "/home/lantern/stats"
```
oxtoacart: A more descriptive name like stats_path may improve readability.

oxtoacart: Also, should we pass this in as a command-line argument?

oxtoacart:
@aranhoide Some unit tests would be helpful to make sure this is working as expected and to help folks like me understand what the code is doing :)

oxtoacart:
@aranhoide If you plan to write some unit tests, I'll wait for those before continuing my review as they'll make it easier for me to unpack the logic.  If not, I'll continue without - just let me know.

lib/stats.py
uaalto: I like this functional technique.

lib/stats.py
aranhoide: I thought the name was kind of namespaced already by virtue of being in a module called `stats`, but I've changed it anyway.

I don't have any need to make the path configurable currently, so I'll call [YAGNI](https://en.wikipedia.org/wiki/You_aren't_gonna_need_it) on this for now.

lib/stats.py
```diff
@@ -27,7 +27,7 @@
 
 delta_0 = delta_val(val=None, prev=None)
 
-avg_0 = avg_val(val=0, nsamples=0)
+avg_0 = avg_val(val=None, nsamples=0)
```
aranhoide: I don't care a whole lot what the value for the case of no samples is, but `None` is probably most informative.

lib/stats.py
```diff
@@ -38,7 +38,7 @@ def delta_reduce_step(accum, newval):
     return delta_val(nextval, newval)
 
 def avg_reduce_step(accum, newval):
-    return avg_val(val=(accum.val * accum.nsamples + newval)
+    return avg_val(val=((accum.val or 0) * accum.nsamples + newval)
```
aranhoide: Allowing for `None`.

lib/stats.py
```diff
@@ -98,32 +98,55 @@ def save():
...
+    while True:
+        line = fileobj.readline()
+        if not line:
+            break
+        s = parse_line(line)
+        if s.time < start_time:
+            continue
+        yield s
+
+def reduce_stats(dimensions, samples):
```
aranhoide: Extracting the algorithmically meaty part of this into an easier-to-test function, which will be hopefully a bit easier to understand without the file access and iteration parts amalgamated in it.

aranhoide:
@oxtoacart : I've isolated a bit more the part that, being more algorithmically complex, is more likely to have bugs or gain them in future updates, and I've added a few tests for that.  I think the other parts are straightforward enough and are validated by the fact that manual tests work at all.

Let me know what other tests you think are needed, if any.

In a sense, `get_bps` and `check_load` are client code that explain what this system is currently used for.

 `get_bps` gives you a single-figure summary of transfer, which is more convenient to fetch through (e.g.) SSH than the raw samples it's calculated from.  It has been used in manually run compaction scripts, where the throughput of hundreds of proxies was needed to decide on a good reallocation of users.

check_load (which will be deleted from here; `salt/http_proxy/check_load.py` will gain similar logic) tries to guess whether a system is overloaded based on recent activity.

The major risk in this module is, I think, `save` failing because of an API change in `psutil` when we upgrade it (most probably as part of an OS upgrade).  I'll add a note in the ticket for upgrading the Ubuntu version.

test/test_stats.py
```diff
@@ -0,0 +1,88 @@
...
+                      {'actual_start_time': self.sample_times[0],
+                       'values': {'load_avg': 1.3, 'disk_tx': 100}})
+
+    def test_three_non_monotonically_increasing_samples(self):
+        """
+        For dimensions of accum_type==delta_type, a value that is lower than
+        the previous one will be taken as evidence that there has been a
+        reboot.
+
+        In this case, we count 50 disk_tx before the reboot and 100 after. This
```
oxtoacart: Very cool!

lib/stats.py
```diff
@@ -0,0 +1,180 @@
...
+
+delta_0 = delta_val(val=None, prev=None)
+
+avg_0 = avg_val(val=None, nsamples=0)
+
+def delta_reduce_step(accum, newval):
+    if accum.val is None:
+        return delta_val(0, newval)
+    nextval = accum.val + newval
+    if newval >= accum.prev:
```
oxtoacart: Might warrant a comment that this is where we exclude reductions in the value which are attributed to a restart.

lib/stats.py
```diff
@@ -0,0 +1,180 @@
...
+    ('disk_pc', 'disk %', avg_type, float),
+    ('disk_tx', 'disk TX', delta_type, int),
+    ('bytes_sent', 'bytes sent', delta_type, int),
+    ('bytes_recv', 'bytes recv', delta_type, int),
+    ('net_errors', 'net errors', delta_type, int),
+    ('net_dropped', 'net dropped', delta_type, int)]]
+
+name2def = {d.name: d for d in stat_defs}
+
+def save():
```
oxtoacart: Comments on these methods would be good (i.e. append a 1 line snapshot of current system stats to the stats file).

lib/stats.py
```diff
@@ -0,0 +1,180 @@
...
+    now = datetime.utcnow()
+    la1m, _, _ = os.getloadavg()
+    net = psutil.net_io_counters(pernic=False)
+    mem = psutil.virtual_memory()
+    swap = psutil.swap_memory()
+    cpu = psutil.cpu_times_percent(interval=1)
+    duse = psutil.disk_usage('/')
+    dtx = psutil.disk_io_counters(perdisk=False)
+    if os.path.exists(stats_path):
+        lines = file(stats_path).readlines()[-num_samples:]
```
oxtoacart: The reason we do this instead of just appending is because we want to cap the size of the file and treat it like a rolling buffer?

oxtoacart:
This looks great!  Thanks for the unit test, that helped me a lot.  Without it, I would have been pretty lost (my level of Python is much below yours).

The only comments remaining are about potentially adding some comments to the code, nothing substantive.  I'm ready to merge whenever you give me the go ahead (or you can merge yourself).

lib/stats.py
aranhoide: Will do, thanks.

lib/stats.py
aranhoide: Will do too.

lib/stats.py
aranhoide: Yes, I'll add a comment on that too.

aranhoide:
@oxtoacart :  Thanks for the review and suggestions!  I've added comments in the places you requested, which includes all the public functions in the module.  I've also made it explicit which are those, by adding an `__all__` (which, in Python, determines what you get if you `import *` from a module, and is thus a semiformal way to express its intended interface).  Let me know if anything is unclear, incomplete or wrong.

oxtoacart:
Nice work!

PR #210, opened 2016-06-06
Update README.md

fffw:
Add instruction to launch a new proxy

### DIFF
--- a/README.md
+++ b/README.md
@@ -227,7 +227,7 @@ salt-cloud -u
 
 ```
 cloudmaster_name = "cm-donyc3myname"
-cloudmaster_address = "188.166.40.244"
+cloudmaster_address = "188.166.40.244" # Must use IP address here. Lots of pillars depend on an correct IP address.
 ```
 
 - and place it in `~/git/lantern_aws/bin/config_overrides.py`.  You probably want to have it saved somewhere else too, since you'll be deleting and restoring `config_overrides.py` to alternate between the production deployment and one or more non-production ones.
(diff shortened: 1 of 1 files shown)


fffw:
@aranhoide I just added my steps to create a new proxy. Please revise if there's anything missing or incorrect.

README.md
```diff
@@ -245,6 +245,32 @@ Your cloudmaster should be ready now.  If it's not a production one (XXX: add in
...
+First, add an request to the queue.
+```python
+from redis_util import redis_shell as r
+r.delete('donyc3myname:srvreqq') # run multiple times until it returns 0
+r.lpush('donyc3myname:srvreqq', -1) # end mark
+r.lpush('donyc3myname:srvreqq', 1) # can be any positive number not already there
+r.lrange('donyc3myname:srvreqq', 0, -1) # to verify, the result should be [1, -1]
+```
+
+Then run the refill service.
```
aranhoide: As of `master`, the `refill_cm_srvq` service should be enabled in all cloudmasters, test or production, so this step shoudn't be necessary.

README.md
```diff
@@ -245,6 +245,32 @@ Your cloudmaster should be ready now.  If it's not a production one (XXX: add in
 
 [2] You can't use `bin/hscloudmaster.bash` here because it hasn't been adapted to work as root, which is only needed during this bootstrap procedure.
 
+##### Launching a proxy
+
+First, add an request to the queue.
+```python
+from redis_util import redis_shell as r
+r.delete('donyc3myname:srvreqq') # run multiple times until it returns 0
```
aranhoide: `delete` deletes the whole queue, so you should never need do it more than once.  If it ever gets refilled, this is probably the local refill_cm_srvq crashing and respawning, which creates the `-1` sentinel value again.

aranhoide: These instructions have the advantage of being robust to whatever is the current state of the queue, but the only step that should be needed in normal usage is adding the new request.

aranhoide:
@fffw : this procedure is probably more work than necessary in most cases, but has the advantage that it should work in all cases, except for one thing: if your cloudmaster is up-to-date, it should already be running a `refill_cm_srvq` service, which will compete with your shell-spawned one to grab the job.  So spawning a shell version of the service should only ever be needed for debugging the process of launching proxies.

fffw:
Thanks for the details review @aranhoide . Updated based on your comments

aranhoide:
LGTM, thanks!

PR #215, opened 2016-06-06
Not restarting redis unnecessarily on highstate

oxtoacart:


oxtoacart:
What was happening is that redis.conf was getting overwritten from the static version every time, then we appended redis_master (or redis_slave), and then the redis-server service said "oh, redis.conf changed, let me restart".

oxtoacart:
Now, I template the redis.conf and file.managed is smart enough not to change anything if the result of evaluating the template is unchanged.

aranhoide:
LGTM!

PR #216, opened 2016-06-06
Fixed redis startup warnings

oxtoacart:


aranhoide:
Sorry, I had missed this one.  Nothing jumps out to me as wrong here, but I have no idea what this does, barring its fixing the warnings.  I'll do some googling and comment on what I find.  @fffw : if you have any experience with this kind of stuff could you please comment on this PR too?

salt/redis_server/init.sls
```diff
@@ -111,6 +111,30 @@ disable-redis-server-sysv:
     - group: root
     - mode: 644
 
+# Allow memory overcommit to make sure Redis background save still works even in
+# low memory conditions.
+vm.overcommit_memory:
```
aranhoide: I assume this is good, since it's recommended here (see second bullet point).

http://redis.io/topics/admin

salt/redis_server/init.sls
```diff
@@ -111,6 +111,30 @@ disable-redis-server-sysv:
...
+  sysctl.present:
+    - value: 1
+
+# Increase somaxconn for Redis
+net.core.somaxconn:
```
aranhoide: Google turned out this for background: https://tweaked.io/guide/kernel/

Definitely better than the default 128!  Given the size of our deployment I don't think it would be crazy to bump this further?  Anyway, LGTM.

salt/redis_server/init.sls
```diff
@@ -111,6 +111,30 @@ disable-redis-server-sysv:
...
+  sysctl.present:
+    - value: 1024
+
+# Disable THP for Redis's benefit. This is done permanently in rc.local
+disable-THP:
```
aranhoide: This page lists this as a "must".  Enough for me.

http://redis.io/topics/latency

salt/redis_server/init.sls
```diff
@@ -122,6 +146,9 @@ redis-server-running:
     - watch:
         - file: /etc/redis/*
         - file: /etc/init/redis-server.conf
+        - file: /etc/rc.local
+        - sysctl: vm.overcommit_memory
+        - sysctl: net.core.somaxconn
```
aranhoide: Will restarting the service do anything for the `sysctl`s?  In the first page referred above a reboot was indicated.  I don't think it's crazy to add a reboot command that watches these instead.

salt/redis_server/rc.local
```diff
@@ -0,0 +1,16 @@
...
+# This script is executed at the end of each multiuser runlevel.
+# Make sure that the script will "exit 0" on success or any other
+# value on error.
+#
+# In order to enable or disable this script just change the execution
+# bits.
+#
+# By default this script does nothing.
+
+echo never > /sys/kernel/mm/transparent_hugepage/enabled
```
aranhoide: +1, see my comment above about THP.

aranhoide:
@oxtoacart LGTM.  Totally out of my depth here, so I don't know whether these changes will all get applied alright.  In particular I'm concerned about the reboot requirement.  But since you've tested this to not break anything and remove warnings, I think we're far better off with this merged.  We can add the rebooting, if necessary, in a separate PR.

oxtoacart:
@aranhoide Thanks for the review!

salt/redis_server/init.sls
fffw: Reboot is not required if `sysctl` command is run. https://docs.saltstack.com/en/latest/ref/states/all/salt.states.sysctl.html#salt.states.sysctl.present says 
> Ensure that the named sysctl value is **set in memory and persisted** to the named configuration file.

oxtoacart: Thanks guys!  I know that the sysctl values take effect on the system immediately, I'm just restarting the process in case it is still governed by the values that were in place when the process itself started.

PR #218, opened 2016-06-07
Add randomization of proxy protocol and port

oxtoacart:
I've tested that manually configuring the proxy_port and obfs4_port has the desired effect after running state.highstate on the proxy.  I've also added the code in refill_srvq.py to randomize the configuration for new proxies, but for the life of me I couldn't figure out how to get something into the queue that the script would read.  @aranhoide do you mind testing that piece of it?

cc: @myleshorton

### DIFF
--- a/salt/cloudmaster/init.sls
+++ b/salt/cloudmaster/init.sls
@@ -67,9 +67,6 @@ refill_srvq:
         - enable: yes
         - require:
           - cmd: {{ svc }}-services-registered
-        - watch:
-          - file: /usr/bin/refill_srvq.py
-
 
 {{ svc }}-services-registered:
     cmd.run:
--- a/salt/cloudmaster/init.sls
+++ b/salt/cloudmaster/init.sls
@@ -66,7 +66,9 @@ refill_srvq:
     service.running:
         - enable: yes
         - require:
-              - cmd: {{ svc }}-services-registered
+          - cmd: {{ svc }}-services-registered
+        - watch:
+          - file: /usr/bin/refill_srvq.py
 
 
 {{ svc }}-services-registered:
--- a/salt/cloudmaster/refill_srvq.py
+++ b/salt/cloudmaster/refill_srvq.py
@@ -135,7 +135,7 @@ def run():
         time.sleep(10)
 
 def new_proxy_name(req):
-    type_str = 'obfs4' if req.get('obfs4_port', 0) > 0 in req else 'https'
+    type_str = 'obfs4' if req.get('obfs4_port', 0) > 0 else 'https'
     return vps_util.new_vps_name('fp-' + type_str)
 
 def launch_one_server(q, reqid, name, req_string):
--- a/salt/cloudmaster/refill_srvq.py
+++ b/salt/cloudmaster/refill_srvq.py
@@ -135,7 +135,7 @@ def run():
         time.sleep(10)
 
 def new_proxy_name(req):
-    type_str = 'obfs4' if 'obfs4_port' in req else 'https'
+    type_str = 'obfs4' if req.get('obfs4_port', 0) > 0 in req else 'https'
     return vps_util.new_vps_name('fp-' + type_str)
 
 def launch_one_server(q, reqid, name, req_string):
--- a/salt/cloudmaster/refill_srvq.py
+++ b/salt/cloudmaster/refill_srvq.py
@@ -6,6 +6,7 @@ from Queue import Empty
 import os
 import time
 import traceback
+import random
 
 import yaml
 import json
@@ -91,7 +92,20 @@ def run():
                 if isinstance(req, int):
                     # Transition: support the old format while we are updating
                     # the config server etc.
-                    req = {'id': req, 'srvq': QPREFIX + ':srvq'}
+                    # We also randomize whether it's obfs4 or not, and which
+                    # port to use.
+                    proxy_port = 443
+                    obfs4_port = 0
+                    if random.random() > 0.5:
+                        # Pick a random high port higher than salt port 4506
+                        proxy_port = random.randint(5000, 60000)
+                    if random.random() > 0.5:
+                        # Make this an obfs4 proxy
+                        obfs4_port = proxy_port
+                    req = { 'id': req,
+                            'srvq': QPREFIX + ':srvq',
+                            'proxy_port': proxy_port,
+                            'obfs4_port': obfs4_port }
                     req_string = json.dumps(req)
                 reqid = req['id']
                 if reqid in pending:
--- a/salt/http_proxy/init.sls
+++ b/salt/http_proxy/init.sls
@@ -99,6 +99,7 @@ save-access-data:
         - user: lantern
         - group: lantern
         - cwd: /home/lantern
+        - order: last
         - require:
             - file: {{ fallback_json_file }}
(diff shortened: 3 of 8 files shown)


salt/http_proxy/init.sls
```diff
@@ -72,6 +72,11 @@ include:
             - file: fp-dirs
 {% endfor %}
 
+allow-bind-low-port:
+  cmd.run:
+    - name: setcap 'cap_net_bind_service=+ep' /home/lantern/http-proxy
```
oxtoacart: Note - instead of doing port redirection, we just allow the http-proxy to listen on the low 443 port.  The reason that we had 62443 open was that it needs to be open for the redirection to work, which stinks.  This way, open only the port that we need.

oxtoacart:
Note, this also takes care of getlantern/lantern-internal#7

oxtoacart:
I did a little testing by just hacking together something fake that bypasses the queue, and from that things look okay.

salt/http_proxy/init.sls
aranhoide: Nice!

salt/cloudmaster/refill_srvq.py
```diff
@@ -91,7 +92,20 @@ def kill_task(reqid):
                 if isinstance(req, int):
                     # Transition: support the old format while we are updating
                     # the config server etc.
-                    req = {'id': req, 'srvq': QPREFIX + ':srvq'}
+                    # We also randomize whether it's obfs4 or not, and which
+                    # port to use.
+                    proxy_port = 443
```
aranhoide: Not worth "fixing", but in Python an if/else is more idiomatic.

salt/cloudmaster/init.sls
```diff
@@ -66,7 +66,9 @@ refill_srvq:
     service.running:
         - enable: yes
         - require:
-              - cmd: {{ svc }}-services-registered
+          - cmd: {{ svc }}-services-registered
+        - watch:
+          - file: /usr/bin/refill_srvq.py
```
aranhoide: It was on purpose that I didn't have the refill service auto restart when the config files changed, because a restart would cause existing proxies to be left abandoned half-configured, so they'd need to be manually destroyed.

It's perhaps not a huge deal because abandoned servers happen anyway for other reasons and we need to watch the VPS checks anyway.

I imagine if we want to watch this we want to watch the upstart .conf files too.

salt/cloudmaster/refill_srvq.py
```diff
@@ -121,7 +135,7 @@ def kill_task(reqid):
         time.sleep(10)
 
 def new_proxy_name(req):
-    type_str = 'obfs4' if 'obfs4_port' in req else 'https'
+    type_str = 'obfs4' if req.get('obfs4_port', 0) > 0 in req else 'https'
```
aranhoide: Good catch!

salt/http_proxy/init.sls
aranhoide: Nice!

salt/http_proxy/init.sls
```diff
@@ -94,6 +99,7 @@ save-access-data:
         - user: lantern
         - group: lantern
         - cwd: /home/lantern
+        - order: last
```
aranhoide: I guess this can't hurt, but why is it needed?

aranhoide:
LGTM, other than I have no idea how the new port permissions work.  I'll look it up, but if you have tested this part then this doesn't need wait for me.

Holding the merge while we figure out your problems for testing this end-to-end.

salt/cloudmaster/init.sls
oxtoacart: Oh, sorry, I hadn't thought about it.  I'll take out the watch.

salt/http_proxy/init.sls
oxtoacart: For obfs4, save_access_data.py depends on a file that's written by the proxy process itself, so we have to wait for that to finish.  I tried doing that with a dependency on the service, but that didn't work for some reason.

aranhoide:
@oxtoacart : I've already said this in Slack, but if what you want is just to launch a new proxy in your test cloudmaster you can follow these instructions:

https://github.com/getlantern/lantern_aws#launching-a-proxy

oxtoacart:
@aranhoide Thanks for the review!  If by port permissions you mean the firewall settings, I tested all that by manually changing the pillars and running highstate.  I tried high and low ports, and tried HTTPS and OBFS4.

salt/http_proxy/init.sls
aranhoide: Sounds like a race condition.  I'm not sure even `- order: last` will guarantee that the file has been written by the proxy process, although I guess anecdotally that's the case often enough if it seems so in your tests.  I guess the proper way to guarantee this is to make a small script that waits until the file is there (and perhaps errors after a timeout).  That script can then depend on the service or go in `- order: last`.

salt/http_proxy/save_access_data.py
```diff
@@ -40,12 +41,21 @@ def add_http_access_data(d_out):
 def add_obfs4_access_data(d_out):
     d_out['pluggabletransport'] = 'obfs4'
     p = re.compile('Bridge obfs4.+cert=([^ ]+) iat-mode=([0-9])')
-    with open('obfs4_bridgeline.txt') as f:
-        for line in f:
-            m = p.match(line)
-            if m:
-                d_out['cert'] = m.group(1)
-                d_out['pluggabletransportsettings'] = {'iat-mode': m.group(2)}
+    for i in range(0, 10):
```
oxtoacart: Now we wait for a while for the file to become available

oxtoacart:
@aranhoide Okay, I think this is ready for final review.

salt/http_proxy/init.sls
oxtoacart: Fixed.

salt/http_proxy/save_access_data.py
```diff
@@ -40,12 +41,21 @@ def add_http_access_data(d_out):
...
+        time.sleep(2)
+        try:
+            with open('obfs4_bridgeline.txt') as f:
+                for line in f:
+                    m = p.match(line)
+                    if m:
+                        d_out['cert'] = m.group(1)
+                        d_out['pluggabletransportsettings'] = {'iat-mode': m.group(2)}
+                return
+        except Exception, e:
```
aranhoide: If we fail the ten times the exception will still be eaten by this and the script will exit with success.  Maybe we should just `raise` it in this case, so it's harder to miss?

salt/http_proxy/save_access_data.py
```diff
@@ -40,12 +41,21 @@ def add_http_access_data(d_out):
...
+            if i == 9:
+                # Last pass, log exception
+                logging.exception(e)
```
aranhoide: I mean, just change this line to `raise e` or just `raise` (in the latter case you don't need to assign a name to the exception).

aranhoide:
LGTM, other than I think we could make it a bit harder to miss if we fail to load or process the 'obfs4_bridgeline.txt'.

salt/http_proxy/save_access_data.py
oxtoacart: Oh yeah, good point.  Fixed.

PR #219, opened 2016-06-08
make sshalert.py ask the ops panel instead of redis

aranhoide:
Not the most important issue but I got enough of a head start on this when redis was down that I thought I'd just go ahead and end this big PITA that is hussh/redis whitelisting, especially in test cloudmasters.

This change makes all VPSs check for whitelisted IPs by calling an endpoint in our ops panel.  If our ops panel is down, the user will be assumed to be whitelisted.

To whitelist yourself in the ops panel, just log into `ops.panel.io` with your github account.  If you are a member of the getlantern organization, your IP will get whitelisted.  Note that you need to give your app permission to see your membership (the membership of some Lantern devs is not public).

If you leave the `ops.panel.io` tab open, it will keep whitelisting your IP even if it changes.  Useful if you're tethering in a train or something, I guess. :)

### DIFF
--- a/bin/update.py
+++ b/bin/update.py
@@ -135,6 +135,8 @@ def upload_pillars(as_root=False):
         # Bypass stunnel in dev environments because we're not encrypting connections to Redis
         redis_via_stunnel_url = cfgsrv_redis_url
 
+    ssh_whitelist_query_token = util.read_ssh_whitelist_query_token()
+
     util.ssh_cloudmaster((
             'echo "salt_version: %s" > salt.sls '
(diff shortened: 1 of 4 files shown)


aranhoide:
Merging since I've tested every use case (even when ops panel is down) and this errs heavily on the side of not bothering you unnecessarily.

fffw:
@aranhoide Just a side note, the ops panel is at `ops.lantern.io` instead of `ops.panel.io`

aranhoide:
@fffw : thanks!  Fixed.

PR #221, opened 2016-06-09
generalize queue/config table fixing utilities, add ad-hoc function t…

aranhoide:
…o fix bad indentation

Indentation is fixed simply by passing the indented part of the config (everything but the server name) through the fixed `serialize_access_data`.

I've tested this in a fake proxy queue and in a fake srv->cfg table, each of which I populated with real broken and already-good entries.

Fixed entries will still be detected as needing a fix, but this is a harmless bug, because I've tested that the fix function is idempotent on the fixed config.  So even if we need to run this more than once for any reason, we're good.


aranhoide:
@oxtoacart : could you take a look at this before I run it?  Fixing queues and the server table will be done by just running `fix_all_bad_indentations`.  I've tested all the parts, but since this has the potential, however unlikely, of breaking our configs in some way or the other, a PR review seems advisable.

aranhoide:
Merging and running since nobody seems available for review right now and there are some 40+ live proxies unusable, besides this breaking checkfallbacks and sanity checks.

lib/vps_util.py
```diff
@@ -441,48 +441,58 @@ def fix_cert_newlines(d):
...
-    return '|'.join([ip, name, serialize_access_data(fixed_access_data, name)])
-
-def filter_queue(qname, fn):
+def fix_queue(qname, fix_fn):
+    # The backup is left around just in case you bork something.
+    # Delete it manually or by calling `delete_q_backups` after
+    # making sure everything went well.
+    print "handling %s..." % qname
     bakname = qname + '.bak'
     redis_shell.rename(qname, bakname)
```
oxtoacart: Clever

oxtoacart:
Code LGTM.  Thanks for fixing this!

PR #228, opened 2016-06-16
Switch from stunnel to tlsproxy

oxtoacart:
cc: @aranhoide @myleshorton 

This is currently deployed to staging and works there.  The tlsproxy code is [here](https://github.com/getlantern/tlsproxy).


bin/util.py
```diff
@@ -89,7 +89,6 @@ def set_secret_permissions():
             os.chmod(os.path.join(path, name), stat.S_IREAD)
 
 def ssh_cloudmaster(cmd=None, out=None, as_root=False):
-    whitelist_ssh()
```
oxtoacart: @aranhoide I took this out since we're no longer using Redis whitelisting.

salt/config_server/init.sls
```diff
@@ -31,9 +31,9 @@ cfgsrv-env-port:
     - pattern: "^PORT=.+$"
     - repl: PORT=62000
 
-/etc/stunnel/stunnel_client.conf:
+/etc/init/tlsproxy-client.conf:
```
oxtoacart: A bonus of using the tlsproxy is that we easily get to manage it with upstart.

salt/redis_tls/init.sls
```diff
@@ -1,11 +1,11 @@
 {% set files=['client_key.pem', 'client_cert.pem'] %}
 
 {% if pillar["in_production"] %}
-{% set files=files + ['redis.getlantern.org/fullchain.pem', 'redis.getlantern.org/cert.pem', 'redis.getlantern.org/privkey.pem'] %}
+{% set files=files + ['redis.getlantern.org/fullchain.pem', 'redis.getlantern.org/chain.pem', 'redis.getlantern.org/cert.pem', 'redis.getlantern.org/privkey.pem'] %}
```
oxtoacart: We need the chain for the Go tls stack.

aranhoide:
@oxtoacart sorry this realization only came now, but:

In top.sls I see stunnel4 is only enabled for the config-server.  So IIRC we're only going to use it for services that talk to redis through Carmine, right?  If that is the case, then we shouldn't need TLS at all because all those (config servers, pro server, ops panel -- when this does get to use redis) are in the same datacenter as the redis box, so they can connect to its IP address in the private network.

The only complication in that would be failover.  But we control IP assignment completely in the private DO network, so we can do IP-based failover: when a slave is promoted to master, they configure themselves to use the master IP.

oxtoacart:
@aranhoide Yes, I was thinking of using floating IPs for auto failover too.

The pro server also talks to Redis, and it's in a different datacenter, so we need this one way or the other.  Plus, I think it's a good idea to use even for the stuff in the same datacenter.

oxtoacart:
@aranhoide Also, the slave Redis instance is in another data center for availability reasons, so in a failover scenario we are actually traversing across data centers.

aranhoide:
@oxtoacart oh yes, I forgot the Pro server is in Heroku.  Now, moving a Clojure app from Heroku to DO is trivial if that's all the problem, esp. if it's not live yet.

I think encrypting traffic in the private network is overkill.  If we're concerned about someone rooting a proxy in the same datacenter or something like that, we can have the Carmine guys and redis in NYC2 and everything else in NYC3.  Both support private networking.

As for failover, I think it's worth moving at least the config servers along with the redis instance.  So we might as well move them all.

oxtoacart:
> As for failover, I think it's worth moving at least the config servers along with the redis instance. So we might as well move them all.

That's in a full datacenter failure.  There are failure scenarios where the only thing that has to move is Redis, and that only temporarily.

> I think encrypting traffic in the private network is overkill.

That depends on your requirements.  When I was working with healthcare data, it was actually required.  Since we have it, it works and it has much reduced overhead now, I say we keep it.

Also, keep in mind that we have all these python scripts from other machines, including developer machines on their home networks, that also have to talk to Redis, and those connections should definitely be encrypted.

aranhoide:
>> As for failover, I think it's worth moving at least the config servers along with the redis instance. So we might as well move them all.

> That's in a full datacenter failure.  There are failure scenarios where the only thing that has to move is Redis, and that only temporarily.

I'd move at least the config servers even if they could keep working otherwise, just for the reduced latency and increased throughput.  As for the others, it's a wash, but the mechanism to move the config servers would probably work for them.

>> I think encrypting traffic in the private network is overkill.

> That depends on your requirements.  When I was working with healthcare data, it was actually required.  Since we have it, it works and it has much reduced overhead now, I say we keep it.

It's a new system to maintain and keep up and running.  An order of magnitude less than being the dominant CPU consumer is still quite some overhead.  Any bugs in this code have still to surface and be found and fixed.

> Also, keep in mind that we have all these python scripts from other machines, including developer machines on their home networks, that also have to talk to Redis, and those connections should definitely be encrypted.

Isn't traffic from Python scripts in general encrypted when they talk to a `rediss://` URL?  Is all traffic from redis-py going through stunnel currently?

aranhoide:
Also, there's little point in having the slave redis in a different datacenter if we're not ready to move the config-server and Pro server from the old one in a general datacenter failure.

oxtoacart:
> Isn't traffic from Python scripts in general encrypted when they talk to a rediss:// URL? Is all traffic from redis-py going through stunnel currently?

Redis doesn't speak TLS, so it needs a stunnel/tlsproxy in front of it.  Clients whose driver is able to speak TLS, like our Python and Go clients, talk directly to the tlsproxy on the redis server.  Clients whose driver doesn't speak TLS, like our Clojure applications, need their own tlsproxy to upgrade connections from the local end to TLS.

```
Python Script -> (tlsproxy -> redis)
(Config Server -> local tlsproxy) -> (tlsproxy -> redis)
```

So, there's no way to avoid having a stunnel-like thing running on the redis server, the only question is whether we get rid of the one running on the config servers.  To do that, we would have to configure them to connect to Redis via an IP on the private network, which is additional work beyond what we have already.  If that's something we want to do, I'd open a separate ticket for that.

> I'd move at least the config servers even if they could keep working otherwise

The slave redis instance is just in another NY data center, so I don't think the latency would be all that bad.  Also, in the case where only the Redis server failed, it would be easier to just set up a new master in DONYC3 and switch back to that once it's ready, rather than moving the config servers.

aranhoide:
Sorry, as we talked in Slack I had misunderstood how the server side of stunnel is used, and missed the fact that proxies will become major redis hogs soon.

In that light, this LGTM.  [spiped](http://redis.io/topics/encryption) is recommended in the redis webpage, but as @oxtoacart says, it's using OS threads so it seems unlikely to scale as well.

aranhoide:
Salt setup looks ready to merge to me.  I guess people that have more recent experience with similar code (the client, http-proxy) should review tlsproxy itself.

oxtoacart:
Thanks @aranhoide!

@uaalto @fffw Here's the code for the tlsproxy - https://github.com/getlantern/tlsproxy/blob/master/tlsproxy.go

If/when you have some time, would you mind taking a quick look?  It's already working okay in staging.

oxtoacart:
FYI, I've deployed this to prod.  So far it's not a clear improvement CPU wise, but I'm keeping an eye on it.

PR #230, opened 2016-06-22
Auto offload mvp

aranhoide:
secret_lib is included as a submodule and everything in secret_lib/salt has the sole purpose of enabling this PR, so it should be considered part of it.  Unfortunately, at some point I started using the PR branch as though it was master, forgetting I was in it, so it became impractical to separate again.  So just review the `salt` folder in the `secret_lib` project in the master branch.

In addition to new unit tests in lantern_aws/test and secret_lib/salt/, I've tested this in my test cloudmaster against a proxy and split server using the following script (barring, I patched the functions that actually detect moderate or severe load to always return True whenever I needed):

```org
* Nothing happens if machine is not overloaded.
* If lightly overloaded
** If recently responded to this, ignore
** If open: close it
*** And store existing slice
** If closed: offload by 10%
* If heavily overloaded
** Try and split it
*** If already closed, try and find existing slice
It will only work if we have assigned the slice to some other server.
*** If not, use current extents
** Offload 50% in any event
* Test offload service for all new cases
** Offload without replacement proxy
* Test split service
```


lib/misc_util.py
```diff
@@ -126,5 +135,34 @@ def confirm(msg):
...
+@contextmanager
+def tempdir(id_):
+    olddir = os.getcwd()
+    workdir = tempfile.mkdtemp(prefix='misc_util_%s_' % id_)
+    os.chdir(workdir)
+    try:
+        yield
+    finally:
+        os.chdir(olddir)
+        os.system("rm -rf " + workdir)
```
oxtoacart: You could use `shutil.rmtree` for this if you wanted.

lib/test_util.py
```diff
@@ -0,0 +1,35 @@
+from contextlib import contextmanager
+import importlib
+import sys
+
+import misc_util
+
+
+@contextmanager
+def mp(mod_name, **modules):
+    "Monkey patched context for tests."
```
oxtoacart: Some more comment on what exactly this does would be useful.

lib/vps_util.py
```diff
@@ -159,7 +159,11 @@ def __init__(self, name, ip, srv):
         self.ip = ip
         self.srv = srv
 
-def actually_offload_proxy(name=None, ip=None, srv=None, pipeline=None):
+def actually_offload_proxy(proportion=0.5, replace=True, name=None, ip=None, srv=None, pipeline=None):
```
oxtoacart: Description of the parameters would be helpful.

lib/vps_util.py
```diff
@@ -170,17 +174,29 @@ def actually_offload_proxy(name=None, ip=None, srv=None, pipeline=None):
...
+               for pip, psrv in redis_shell.hgetall(client_table_key).iteritems()
+               if psrv == packed_srv]
+    if proportion == 1.0:
+        tomove = clients
+    else:
+        # Take a random sample so we're not biased towards offloading at either
+        # end of the IP space.  Thus, if this is a splitting server we don't
+        # overload one of the halves more than the other.
+        tomove = random.sample(clients, int(round(proportion * len(clients))))
+    rshell = pipeline or redis_shell
```
oxtoacart: What is pipeline?

lib/vps_util.py
```diff
@@ -170,17 +174,29 @@ def actually_offload_proxy(name=None, ip=None, srv=None, pipeline=None):
...
+    if replace:
+        dest = pull_from_srvq(region)
+        # It's still possible that we'll crash or get rebooted here, so the
+        # destination server will be left empty. The next closed proxy compaction
+        # job will find this proxy and assign some users to it or mark it for
+        # retirement.
+        dest_psrv = redis_util.pack_srv(dest.srv)
```
oxtoacart: redis_util uses redis_shell under the covers. Is that a problem if pipeline is supposed to be used as rshell?

lib/misc_util.py
aranhoide: Thanks!  Will do.

lib/test_util.py
aranhoide: OK, I'll give it a try.

oxtoacart:
@aranhoide Sorry I didn't finish this yesterday, I am getting back to it today.

lib/vps_util.py
aranhoide: Coming in a PR near you soon(TM). :)

lib/vps_util.py
aranhoide: [This](https://github.com/andymccurdy/redis-py#pipelines).  In short, an object that has the same interface as redis_shell but stores all the ops to perform them transactionally as soon as you call execute().  I've added a short comment about that.

lib/vps_util.py
aranhoide: No, it is by design that they have the same interface.

aranhoide:
@oxtoacart no worries, it's a big one.  I've added the comments you asked for.  Let me know if anything seems missing or unclear.

aranhoide:
@oxtoacart : the point of the testing hacks is that there is code that is mostly side effects without a lot of complicated logic to it.  So the things to catch there are silly details like some function expects a tuple but JSON serialization/deserialization has converted the data to a list.  I guess in Java we'd use DI to create mocks around the dependencies of this code for testing, but here I'd rather make the tests more tricky and dirty and keep the tested code straightforward and uncomplicated by these considerations.

There is other code that has some algorithmic meat to it, like `stats.summary` or `sl_util._split` and is worth factoring out and testing as pure functions.

aranhoide:
Also, unit tests are by no means complete, because I soon realized that testing everything at the level of detail of test_check_load would take way longer than the original PR.  At some point I've just resorted to manual end-to-end testing.  I'll just keep adding tests as the code needs to change, and as the arsenal of testing tools keeps building up.

.gitmodules
```diff
@@ -1,3 +1,6 @@
 [submodule "secret"]
 	path = secret
 	url = git@github.com:getlantern/too-many-secrets.git
+[submodule "secret_lib"]
+	path = secret_lib
+	url = git@github.com:getlantern/secret_lib
```
oxtoacart: We don't need to do anything as part of this PR, but this sort of stuff makes me wonder if we should just make lantern_aws private.

lib/test_util.py
```diff
@@ -0,0 +1,74 @@
+from contextlib import contextmanager
+import importlib
+import sys
+
+import misc_util
+
+
+@contextmanager
+def mp(mod_name, **modules):
+    """
```
oxtoacart: Nice comment!

lib/vps_util.py
oxtoacart: Again, very nice comment!

lib/vps_util.py
oxtoacart: Cool!

salt/http_proxy/check_load.py
```diff
@@ -1,94 +1,104 @@
...
+
+def moderately_loaded(lavg, cpu_io, mem_pc, cpus):
+    return lavg / cpus > 1.5 or (cpu_io > 1 and mem_pc > 85)
+
+def respond_severely(msg):
+    # This splits this proxy if it happens to be open, or whichever proxy took
+    # its place if the original slice remains unsplit.
+    util.offload_server(msg, 0.5, split=True)
+
+def respond_moderately(msg):
```
oxtoacart: Since we run check_load.py every minute, it seems unlikely that the load would have changed much after calling close_server and so we will pretty much always fall into offload_server, right?

salt/offload/offload.py
```diff
@@ -0,0 +1,54 @@
...
+        task, remover = q.next_job()
+        if task:
+            task = json.loads(task)
+            print "Offloading %s users from %s (%s); replace: %s, retire: %s" % (
+                misc_util.percent(task['proportion']),
+                task['name'],
+                task['ip'],
+                task['replace'],
+                task['retire'])
+            txn = redis_shell.pipeline()
```
oxtoacart: Since pipeline doesn't execute any commands until the very end, that means our logic during the transaction can't rely on side effects of prior Redis calls being readable yet.  Does that assumption hold true?

test/test_check_load.py
```diff
@@ -0,0 +1,146 @@
...
+    def test_severe_response(self):
+        "When severely loaded, a proxy should offload half its users and try and split its slice."
+        util = logger('util')
+        in_mp(stats=stats_with_summary_that_should_trigger_severe_response,
+              util=util)
+        self.assertEqual(len(util._log), 1)
+        self.assertRegexpMatches(util._log[0], r"util.offload_server\('have severe load \(.+\)', 0\.5, split=True\)")
+
+if __name__ == '__main__':
+    # Let the monkey patching festival commence...
```
oxtoacart: LOL!

test/test_http_proxy_util.py
```diff
@@ -0,0 +1,76 @@
+#!/usr/bin/env python
```
oxtoacart: I can't get this test to run.  I get errors like [these](https://gist.github.com/oxtoacart/33cfe98eeb4ad777967667b72e0e029b).

requirements.txt
```diff
@@ -6,3 +6,4 @@ transit-python==0.8.282
 vultr==0.1.2
 linode-python==1.1.1
 subprocess32
+psutil
```
oxtoacart: I added this missing requirement.

oxtoacart:
@aranhoide I finished another pass at this.  Generally looking good except for that one failing unit test.

.gitmodules
aranhoide: The amount of secret logic is very small currently compared to the rest of lantern_aws and I don't think it's likely to grow, so it doesn't seem like a pressing problem.  OTOH, I don't think anyone else is benefiting much from having lantern_aws public, so I don't have much of a preference for either option.

salt/http_proxy/check_load.py
aranhoide: This is what I'm doing to address that problem: after any response is taken, any other responses to conditions or equal or lower severity are suppressed for a length of time.  That's what "moratory"[1] means in that source file and what's tested [here](https://github.com/getlantern/lantern_aws/blob/auto-offload-mvp/test/test_check_load.py#L77).

[1] The spelling corrector flags this word so a better one should probably be chosen.  Suggestions welcome!

salt/offload/offload.py
aranhoide: The first sentence is correct, although I'd add one nuance: we can't rely on the side effects of prior calls _in the pipeline_ being readable yet.  Redis calls made through the `redis_shell` will get executed immediately.

I'm not sure if you're asking whether you got the pipeline semantics right or whether we're taking that into account, but the answer to both would be yes AFAIK.

test/test_http_proxy_util.py
aranhoide: @oxtoacart: I think you have `lantern_aws/bin` in your `PYTHONPATH` and therefore you get `lantern_aws/bin/util.py` imported instead of `lantern_aws/salt/http_proxy/util.py`.

The only reason for having `lantern_aws/bin` in the PYTHONPATH used to be `hussh`, but now that we aren't using that any more the proper fix for this should be just removing that from your shell init script.

aranhoide: Wait, I think I can make this test resistant to that by a bit more trickery, so I might as well.  I'll push a commit in a sec..

aranhoide: @oxtoacart done; I've tested that adding lantern_aws/bin to my PYTHONPATH gets me the errors you were having, and that my latest commit fixes those.

requirements.txt
aranhoide: Maybe we should mock psutil out in tests instead, because the version in Ubuntu 14.04 is probably incompatible with whatever we're getting in recent Ubuntu or OS X.  Then again, if tests pass with this then it's fine.

aranhoide:
@oxtoacart fixed the failing unit test.  I remind you that the master branch of secret_lib/salt/ is a core part of this feature.  You can send any comments in mail to me if it's too inconvenient to comment on a non-PR.

oxtoacart:
@aranhoide Tests are passing for me.  Just taking a look at the code in secret_lib.

oxtoacart:
@aranhoide I took a look at sl_util.py in secret_lib, and nothing seems obviously wrong.  That said, I don't think I understand what's going on very well, so my review isn't necessarily the most useful.

A general observation I have is that we seem to be relying a lot on somewhat ad-hoc text-based data formats and string manipulation of data in redis.  It might be nice to:

1. Encapsulate the data types as actual Python types and use those wherever we read and write the data

2. Consider switching to standard encodings like Json wherever possible.

[this line](https://github.com/getlantern/secret_lib/blob/master/salt/sl_util.py#L58) is an example of what I mean.  It would be nice to have a Server datatype with a status field of "empty, locked, etc."  It could also encapsulate how to (un)marshal itself to/from a text representation.

aranhoide:
On Wed, Jun 29, 2016 at 4:40 AM, oxtoacart <notifications@github.com> wrote:

> @aranhoide <https://github.com/aranhoide> I took a look at sl_util.py in
> secret_lib, and nothing seems obviously wrong. That said, I don't think I
> understand what's going on very well, so my review isn't necessarily the
> most useful.
>
Hopefully my previous email provides some useful context.  Sorry it had
been missing; I'm not sure where that should be documented, actually;
suggestions welcome.

> A general observation I have is that we seem to be relying a lot on
> somewhat ad-hoc text-based data formats and string manipulation of data in
> redis.
>
This is true currently, but we're moving away from that and standardizing
on JSON for new stuff.

> It might be nice to:
>
>    1.
>
>    Encapsulate the data types as actual Python types and use those
>    wherever we read and write the data
>
> I do that whenever that feels useful, e.g. if there's useful logic to
attach to those types, or if the structure is complex enough, or if
invariants between the parts must be maintained, etc.

But there's a burden of proof to that.  In general, I'm against
encapsulating simple data in "data classes" because that's often just
boilerplate to maintain and it loses all Python standard facilities for
working with lists/tuples/sets/dicts (and also expectations and knowledge
about how those data structures work).  There's some useful middle ground
in named tuples, which I use extensively in lantern_aws/lib/stats.py, but
there naming the fields feels more useful, while here it's just start, end,
and proxy so even that seemed overkill.  A comment is probably missing
somewhere, but I think that's it.

As for serialization, I think the low hanging fruit would be having utility
functions to read/write those data types.  Again, often just json.load and
dump are all that is needed.

>
>    1.
>
>    Consider switching to standard encodings like Json wherever possible.
>
> Yes, as I said that's current policy going forward, but I'm being
parsimonious about "fixing" existing stuff if it involves changes to
several projects (esp. if they involve Lua scripts, which are currently
expensive to test).  Perhaps as we accrue more automated testing it'll make
more and more sense to keep everything tight and up-to-date with current
standards.

>
>
> this line
> <https://github.com/getlantern/secret_lib/blob/master/salt/sl_util.py#L58>
> is an example of what I mean. It would be nice to have a Server datatype
> with a status field of "empty, locked, etc.  It could also encapsulate how
> to (un)marshal itself to/from a text representation.
>
I think a collection of functions like the one containing that line is more
workable than a class, if that's what you mean by datatype.

I actually have a Python class for slices in secret_lib, which I use in
other contexts:

https://github.com/getlantern/secret_lib/blob/master/ipslice.py

And some deserialization utility for fetching all of these for a given
region:

https://github.com/getlantern/secret_lib/blob/master/secret_lib.py#L36

(I don't think the above belongs in the class, BTW).

As for serializing, I haven't found a useful way to encapsulate that that
helps more than it obfuscates.  The slice table is represented as a sorted
set.  "Serializing" it means adding a key, score pair.  Adding or updating
one looks different than adding or removing a bunch.

Anyway, I didn't use this class here because I didn't need any of the
functionality, so the extra (de)serialization steps and dependency wouldn't
net us anything.

All in all, I think these's a big part of stylistic preferences here for
which we can't prescribe a solution in generic terms.  See

http://ravimohan.blogspot.com.es/2007/04/learning-from-sudoku-solvers.html

The best Python I've ever read is in this course[1], and I don't recall it
having used a single class ever.

[1] https://www.udacity.com/course/design-of-computer-programs--cs212

oxtoacart:
Cool that we're doing new stuff in JSON.

I think encapsulating data formats using functions is fine too, I was mostly just saying that it would be good to have it encapsulated in some reusable form.

Thanks for the background on what we're doing.  I'm going to read that and then walk through secret_lib again.

PR #231, opened 2016-06-22
Change throttlethreshold and fix redis init script issue

oxtoacart:


myleshorton:
I feel like this is already deployed @oxtoacart is that right?

oxtoacart:
@myleshorton Only deployed to staging.

salt/http_proxy/config.ini
```diff
@@ -18,7 +18,8 @@ redisclientpk = /usr/secret/redis/client_key.pem
 redisclientcert = /usr/secret/redis/client_cert.pem
 {% if pillar["in_staging"] %}
 enablereports = true
-throttlethreshold = 536870912
+enablepro = true
```
aranhoide: I assume enabling Pro is intended.  Only commenting because it's not in the PR title.

salt/redis_server/redis-server.conf
```diff
@@ -3,8 +3,8 @@
 description "redis server"
 author "Ox Cart <ox@getlantern.org>"
 
-start on (started tlsproxy_server and started tlsproxy_client)
-stop on (stopping tlsproxy_server or stopping tlsproxy_client)
+start on (started tlsproxy-server and started tlsproxy-client)
+stop on (stopping tlsproxy-server or stopping tlsproxy-client)
```
aranhoide: What's the motivation to link these at all?  Just to not duplicate the runlevel definition?  And why this direction instead of the other one?  A redis without tlsproxy makes more sense than a tlsproxy without redis.  Are there any startup dependencies between these?  If so, a comment would be helpful.

Won't we ever want to restart tlsproxy-* without restarting redis-server too?  E.g. to limit downtime, esp. if we ever have any services in the local network talk to redis directly without TLS.

None of this is of much importance and fixing the typo is undoubtedly for the better, so I'll merge anyway.

aranhoide:
LGTM

salt/http_proxy/config.ini
oxtoacart: Yup.

salt/redis_server/redis-server.conf
oxtoacart: Great questions.  The idea was that the slave replicates with the master via tlsproxy, so we don't want the slave trying to connect until tlsproxy is up.

oxtoacart:
@aranhoide Thanks for the review.

PR #232, opened 2016-06-23
Update transfer plot to new DB Schema

uaalto:
It also introduces the concept of sampling through a command-line flag, because now it takes forever to download all keys.


etc/device_tx_analysis.py
```diff
@@ -56,9 +60,28 @@ def main():
...
+        keys = redis_shell.keys('_client:*')
+        numKeys = len(keys)
+        if numKeys == 0:
+            print "There is no bandwidth tracking data. Exiting."
+            sys.exit()
+
+        print "Retrieving transfer data from Redis (%d total keys)..." % numKeys
+        data = []
+        i = 0
+        while i < numKeys:
```
aranhoide: `for i in xrange(numKeys)` would be more idiomatic.  Then you don't need to initialize or increment i manually.

etc/device_tx_analysis.py
```diff
@@ -56,9 +60,28 @@ def main():
...
+            if i % 100 == 0:
+                print "Retrieved %d/%d keys at a sampling rate of %f" % (i, numKeys, sample)
+            if random.random() <= sample:
+                inBytes, outBytes = redis_shell.hmget(keys[i], 'bytesIn', 'bytesOut')
+                data.append([int(inBytes) + int(outBytes)])
```
aranhoide: I don't know if this is what you intend, but this is changing the form of  `data` substantially.  It used to be of the form `[client, bytes, client2, bytes2, ...]` and now it's `[[bytes], [bytes2], ...]`.  Notice the double list nesting; that is quite surprising.  Then again, if the output '.csv' works fine for you then I suppose this must be harmless at worst?

aranhoide:
@uaalto I suppose you wanted me to review this for Pythonic style.  Since this is manually triggered code and I suppose it worked fine for your purposes, I'll merge it.  If you change something according to my comments feel free to push it directly to master unless you want me to review the changes.

etc/device_tx_analysis.py
uaalto: Nice one! Thanks! I actually wanted to find something like this. So I'll change and commit directly.

etc/device_tx_analysis.py
uaalto: I think before it was `[[client, bytes], [client2, bytes2]]`, the reason is that the CSV serializer needs an array of arrays, each nested array being the fields in a line of CSV.

Issue #182, opened 2016-05-25
Comprehensive logging of all proxy lifecycle events


aranhoide:
Events to log:

- launching,
- pulling from queue into service,
- closing *,
- retiring *,
- destroying *.

* log (i) whenever a task is enqueued and (ii) when it's actually performed.  Also provide a reason for the op.

Events should be logged to redis in the same transaction where they happen.  Some service consumes the redis queue and pushes the events forward to borda.  We could also log directly to borda the attempt to perform those ops, so we get a record of failed ops.

aranhoide:
Currently IIRC only closing and retiring are logged, and only when performed automatically, not when run manually.

aranhoide:
The reason for reporting to redis is that some ops are performed atomically in Lua scripts that wouldn't be able to talk to borda directly.  If we are to log ops if and when they are performed, they need to be in the same transaction, be that a Lua script or a MULTI/EXEC op.

aranhoide:
I'll squeeze in the move from Transit to plain JSON too, since Lua scripts can't use Transit either.

oxtoacart:
@aranhoide @myleshorton In answer to the question from our email thread ...

As I understand it, the purpose of this ticket is to provide an automated audit trail of activities that have happened in production.  Borda wouldn't be a good fit for this, and using something like Datomic here makes sense to me.

myleshorton:
Sounds like a plan guys.

aranhoide:
A more comprehensive list of what events to log and what data is needed for each of them:

```org
***** launching:
****** op: launch
****** name
****** cloudmaster
this gets actually stored in the proxy entry, not in the event entry
***** add to (regional, cm) proxy queue
****** op: enqueue
****** name
****** public-ipv4
****** queue
***** pulling from (regional, cm) proxy queue
We can possibly do this atomically with some of the following ones if it's in a
Lua script, but we can also perform this on its own or for baked-in proxies, so
let's leave it separate.
****** op: pull
****** name
****** public-ipv4
****** queue
****** purpose free|pro|baked-in|...
***** putting in srv->cfg
****** op: make-free
****** name
****** public-ipv4
****** srv
***** putting in server->config
****** op: make-pro
****** name
****** public-ipv4
***** adding to slice table
****** op: assign-slices
****** name
****** public-ipv4
****** srv
****** slices
***** closing
****** op: close
****** name
****** public-ipv4
****** srv
***** enqueue for splitting
****** op: enqueue4split
[...]

PR #234, opened 2016-06-27
log2redis: ditch transit in favor of plain JSON

aranhoide:
This is a prerequisite for being able to log some proxy lifecycle events (namely, those that happen inside Lua scripts, where Transit is not available) atomically with their enqueuing/execution, and hence of #182.  While I was at it I made a small interface improvement.  See commit comments.

I thought I'd have this reviewed before going on to add all actual logging, so we avoid having to change all the logs if code review suggests further interface changes.


lib/vps_util.py
```diff
@@ -374,12 +374,7 @@ def retire_proxy(name=None, ip=None, srv=None, reason='failed checkfallbacks', p
     else:
         qname = '%s:retireq' % cm_by_name(name)
     p.rpush(qname, '%s|%s' % (name, ip))
-    log2redis({'op': 'retire',
-               'name': name,
-               'ip': ip,
-               'srv': srv,
-               'reason': reason},
-              pipeline=p)
+    log2redis(p, op='retire', name=name, ip=ip, srv=srv, reason=reason)
```
uaalto: Interesting new syntax. I've checked that it actually builds a dictionary.

lib/redis_util.py
```diff
@@ -24,10 +25,9 @@
                                               'redis',
                                               'client_cert.pem')
 
-redis_shell = redis.from_url(os.getenv('REDIS_URL'), **redis_args)
+redis_shell = redis.from_url(redis_url, **redis_args)
```
uaalto: Better of course.

lib/redis_util.py
```diff
@@ -48,12 +48,9 @@ def log2redis(data, pipeline=None, time=None):
...
     """
-    time = time or utcnow()
-    version = redis_shell.get(versionkey) or 0
-    msg = {'time': time,
-           'version': version,
-           'data': data}
-    s = transit_dumps(msg)
+    if 'time' not in data:
+        data['time'] = str(utcnow())
+    s = json.dumps(data)
```
uaalto: So, this is the gist of the PR. What is the motivation to dump transit? (Besides being less standard than the JSON lib)

aranhoide: The deciding factor is no Transit libraries for Lua, so we couldn't otherwise log things from Lua scripts, like in the config-server PR that depends on this.

uaalto:
LGTM!

PR #240, opened 2016-07-01
increase linode proxy size

aranhoide:
As of this writing, 2GB is the smallest plan offered by Linode.


aranhoide:
Merging this since the change is straightforward, this is urgent to fix, and it cannot be any more broken.

PR #241, opened 2016-07-01
disable launching of new obfs4 proxies, fixes #235

### LINKED ISSUE #235: Stop launching new obfs4 proxies

aranhoide:
Although the change is trivial, I've tested it in my test cloudmaster by launching a proxy.  As of this writing, the launch process is already past the point of this change and going alright.


oxtoacart:
LGTM!

